PR #3, opened 2015-02-24
Reorganized Project Structure

akallem:
I replaced make with ant for our build process, which should make it easier for us to integrate the build process with testing suites or any other tools as needed, and also let each of us work in our own environments as we prefer (e.g., Eclipse, not Eclipse).

I also added a .gitignore.

Last, this includes the initial UI skeleton that we discussed.


src/WordsUI.java
```diff
@@ -0,0 +1,138 @@
...
+			super.paintComponent(g);
+			
+			Graphics2D g2 = (Graphics2D) g;
+			
+			// Background region
+			int x1,y1,x2,y2;			// Boundaries of the board
+			int cx,cy;					// Center of the board
+			
+			x1 = (windowWidth - boardSize)/2;
+			y1 = 20;
```
alexthemark: What is this hardcoded number?

src/WordsUI.java
```diff
@@ -0,0 +1,138 @@
...
+	WordsUI() {
+		window = new JFrame();
+		panel = new JPanel();
+		grid = new Grid();
+		buttons = new Buttons();
+		
+		windowWidth = 750;
+		windowHeight = 750;
+		boardSize = 650;
+		numCells = 20;
```
alexthemark: Is this the same '20' as y1? Can we make this a named constant?

src/WordsUI.java
akallem: Padding at the top of the window before the board is shown.  Will push an update with clarification.

src/WordsUI.java
akallem: Nope, it's different.  Will push an update to clarify.

alexthemark:
lgtm now

PR #4, opened 2015-02-28
First parser

alexthemark:
Creates Command objects that can execute a function on an object, or can create a new object. 

```
<object_name> is a <class> at cell <int, int> .
Make <object_name> <function_name> <function_args> .
```

### DIFF
--- a/src/Command.java
+++ b/src/Command.java
@@ -1,4 +1,20 @@
+import java.util.HashMap;
+
 
 public class Command {
-	//Oh Fuck
+	private CommandType type;
+	private HashMap<String, Object> parameters;
+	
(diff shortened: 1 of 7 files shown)


akallem:
Cool!  I just tried it and it worked for me.  I think the grammar will obviously need to become more structured and generic as we flesh things out, but just to get a first version up and running, this is great.

I'm OK to merge this.

alexthemark:
I think a critical thing to note is that currently the grammar only works line by line... I think this will make things simple for us on this sprint, but it is certainly something that we have to change. I've documented this in #6

alexthemark:
Merging this in so I can better read the diff in your PR.

PR #5, opened 2015-03-01
Gui improvements

akallem:
The GUI can now primitively render objects.  It'll need further refinement, but it should serve us well enough for our work now.  The GUI class also now includes the necessary methods for the back-end of the compiler to manage the view in each frame loop.

### DIFF
--- a/src/WordsUI.java
+++ b/src/WordsUI.java
@@ -51,12 +51,12 @@ public class WordsUI {
 	}
 	
 	/**
-	 * Convert an xy coordinate into a string suitable to serve as a hash key.
+	 * Convert a position into a string suitable to serve as a hash key.
 	 * 
 	 * @return a string representing the coordinate
 	 */
-	private String xyToKey(int x, int y) {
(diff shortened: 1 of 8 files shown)


src/WordsUI.java
```diff
@@ -137,4 +223,59 @@ public void actionPerformed(ActionEvent e) {
...
+	/**
+	 * Adds an object to the UI at a given position.
+	 * 
+	 * @param x The x coordinate of the cell to place the object
+	 * @param y The y coordinate of the cell to place the object
+	 * @param className The name of the class of this object
+	 * @param objName The name of the object
+	 * @param message The message that the object should say.  May be null.
+	 */	
+	public void add(int x, int y, String className, String objName, String message) {
```
alexthemark: We have a class `WordsPosition` that contains x and y. Can we use that in this interface? I think it'll be cleaner: for each object, we pass in the objects that it holds.

akallem: Good idea.  Will do that now.

PR #12, opened 2015-03-12
Complete Lex and Yacc Grammar with AST

akallem:
This branch contains a complete grammar in Lex and Yacc based on the current definition of the language (the examples in the Tutorial on Google Docs) and builds an abstract syntax tree.  It also includes a REPL interface.  For the moment, the tree is dumped to the console so you can see it.

I've added a program called syntax.w in the test directory.  This file contains all of the language constructs listed in the Tutorial file right now.  That means it's not actually a runnable program, since objects are used before they're declared, etc.  The purpose is just to test the parser and AST generation on every language construct before we actually implement them.

It'll print out the AST to the console.  The naming of the nodes should be pretty intuitive.  It doesn't actually evaluate/execute the tree yet, it just constructs the tree and prints it.

I've included some basic syntax error highlighting.  If you make a syntax error, it'll print out the line number and the token in the line near where the error occurred, then skip ahead to the next statement to continue parsing.

The only slightly quirky thing is that I have the lexer recognize the apostrophe s construct rather than the parser.  In other words, the string "John's" is a provided to the parser as a single token (called REFERENCE) rather than two tokens (say, IDENTIFIER followed by APOSTROPHE_S) that the parser would parse.  Without going into detail, the reason is that if I didn't do it this way, it would be difficult to cleanly support the multiple assignment construct "Make Andrew's friend's brother's height 150, brother John, property 100." among other things.  It also has the benefit of ensuring that there's no whitespace between the identifier and the apostrophe s, which I think is a good thing.


alexthemark:
Just gave it a first look over -- this looks beautiful. Will look over further this break.

akallem:
I've added a REPL interface.  Now, it takes a command line argument with the filename, after which a REPL interface is presented.  Try running it with "java -jar jar/Words.jar test/syntax.w".  (Notice there is no redirection to standard input.  It reads the file directly now.)

src/INode.java
```diff
@@ -0,0 +1,68 @@
...
+		} else {
+			indent(level + 1);
+			System.err.println("  empty");
+		}		
+	}
+	
+	@Override
+	public String toString() {
+		String s;
+		String c = "";
```
alexthemark: Should be a `StringBuffer` instead of a `String`. A string is a set length of memory in Java, so a new length of memory will need to be allocated and all chars moved each time you concatenate strings. A StringBuilder, on which you call `.toString()` at the end, doesn't need to copy each character on concatenation.

Not strictly a necessary change (since this is fully working and I doubt we're being graded at all on our Java String efficiency), but I figure it's best if we give each other additional feedback on our code so that we can each improve our Java programming.

src/Words.y
```diff
@@ -13,116 +13,401 @@
...
+	|	value_expression GEQ value_expression		{ $$ = new INode(AST.Type.GEQ, $1, $3); }
+	;
+
+value_expression:
+		reference_list identifier					{ $$ = new INode(AST.Type.RETRIEVE_PROPERTY, $1, $2); }
+	|	literal										{ $$ = $1; }
+	|	NOTHING										{ $$ = new LNode(AST.Type.NOTHING); }
+	|	'(' value_expression ')'					{ $$ = $2; }
+	|	'-' value_expression %prec UMINUS			{ $$ = new INode(AST.Type.NEGATE, $2); }
+	|	value_expression '+' value_expression		{ $$ = new INode(AST.Type.ADD, $1, $3); }
```
alexthemark, line 223: The value of creating the AST here, instead of immediately adding, is so we can later decide if we're doing string concatenation or numerical addition (based on the rules you put in the reference manual), right? 

If so, what is the value add of having multiplication, division, subtraction, and exponentiation handled in a fully formed AST instead of using YACC to simplify them? I want to make sure that we're using YACC to its full power. Do we have specific reasons to build all branches of the AST, and not simplify where we may be able to?

src/Words.y
```diff
@@ -13,116 +13,401 @@
...
 
+public Words(Reader r) {
+	lexer = new Yylex(r, this);
+}
 
-  public Words(Reader r) {
-    lexer = new Yylex(r, this);
-  }
 
+public static Game game;
```
alexthemark, line 321: I don't think this line is necessary anymore. 

Edit: Nvm, but it will be when we have the frame loop enqueue the AST for evaluation

src/Words.y
```diff
@@ -3,7 +3,7 @@
  * All rights reserved.                                                    *
  *                                                                         *
  * This is a modified version of the example from                          *
- *   http://www.lincom-asg.com/~rjamison/byacc/                            *
+ *   http://www.lincom_asg.com/~rjamison/byacc/                            *
```
alexthemark: I don't think we really need this copyright any more... we're a very different file from the lex/yacc calc.

src/Words.y
akallem, line 223: The reason to do the arithmetic in the AST and not Yacc is type checking and coercion.  Since we have dynamic typing and the identifier could refer to anything (number, string, object), we have to build the AST and do the operand type checking (and coercion) during evaluation.

I also think it's much cleaner this way, as it provides a true separation of concerns.  The parser builds the AST, but it has no role in evaluation.

src/Words.y
akallem, line 321: Exactly.

src/Words.y
akallem: Ha, agreed!  I'll remove it now.

src/INode.java
akallem: Cool, great tip.  Just pushed a commit to fix this.

alexthemark:
I'm good with a merge to master.

PR #13, opened 2015-03-24
Grammar additions

akallem:
This branch includes grammar/AST support for the Reset and Exit statements, the touches predicate, predicates using the name of a class (instead of a named object), and "aliases" for the member of the class that satisfied the predicate.

The Exit and Reset statements are internal nodes in the AST in case we want to add parameters to them (like resetting only the listeners, or something like that).  If not, the fact that they're internal nodes with no children is harmless.

The AST for the basic action predicates might need to be revised at some point so it's easier to evaluate, but for now this at least parses the construct.


test/syntax.w
```diff
@@ -106,4 +106,23 @@ Whenever Fred moves left {
 
 Whenever Fred waits {
 	Make John say "hurry up" now. 
-}
\ No newline at end of file
+}
+
+Whenever Mary's dog touches dog {
```
alexthemark: "dog" in this context would be an object named "dog"? since it's not touching "a dog"?

This comment isn't appearing correctly on the "conversation" page but looks right on the "Files changed" page.

akallem: Whoops, that should've been Mary, not dog.  Will change it now.

PR #14, opened 2015-03-27
Skeleton

alexthemark:
Not at all ready for a merge, but putting in a PR so we can see a comparison. Some functions don't exist, others return null values, nothing is tested, etc.

We started working out custom functions, and how these would take parameters. To do this, the AST can take a params hash map in addition to the environment. This method makes conceptual sense: the AST is evaluating its nodes on both the environment that exists, as well as any additional parameters that may exist from the words source code.

Objects can then be held in a symbol table. Classes can be kept global, in a class table. Custom functions can be held by classes.


alexthemark:
Created #16 to fix Say.

PR #32, opened 2015-03-31
Exceptions

### LINKED ISSUE #17: Create Appropriate exceptions
All exceptions subclass WordsException

-WordsClassAlreadyExist
-WordsClassNotFound
-WordsObjectAlreadyExists
-WordsObjectNotFound
-InvalidType
-Reference
-WordsFunctionArg

Exceptions are caught at the statement level of inode.

alexthemark:
A number of exceptions have been created to be thrown from AST evaluation. 
AST Nodes now hold the line number that the node was parsed from. 
Existing inode eval functions now throw appropriate exceptions.
Potential flaw is that these exceptions require a line number, which means that they can only be thrown by a node. 

Closes #17.


mzbenami:
looks nice

alexthemark:
I thought the same thing about your new avatar.

akallem:
I think we do need to be able to throw these exceptions not just by an AST node but also in other places.  For example, in the createObject() method in WordsEnvironment, we need to be able to throw exceptions.

Now, we could handle that by having that method return different kinds of error values, and letting the AST node actually throw the exception.  But that's another layer of infrastructure (error values).

Alternatively, how about making the currentLineNumber a field in WordsEnvironment.  It would be set at the top of eval() before the big switch statement.  This way, the environment knows what line is currently being processed, so anything who has access to the environment (which is anything that matters) can throw an exception referencing that line number.

src/INode.java
```diff
@@ -204,12 +204,20 @@ private ASTValue evalCreateClass(WordsEnvironment environment) {
...
 	}
 
-	private ASTValue evalCreateObj(WordsEnvironment environment) {
+	private ASTValue evalCreateObj(WordsEnvironment environment) throws WordsException {
 		ASTValue objName = children.get(0).eval(environment);
 		ASTValue className = children.get(1).eval(environment);
 		ASTValue properties = children.get(2) != null ? children.get(2).eval(environment) : null;
 		ASTValue position = children.get(3).eval(environment);
 
+		if (environment.getObject(objName.stringValue) != null) {
```
akallem: Per my comment, I kind of like it if this line and the following stayed within the environment, and the environment handles the checks whether the creation is valid.  It separated the concerns better in case we change some things about how objects are managed.  eval just needs to call createObject, and the environment handles the rest.

src/INode.java
```diff
@@ -348,27 +364,30 @@ private ASTValue evalQueueAssignPropertyList(WordsEnvironment environment) {
...
+			if (property.type != WordsProperty.PropertyType.OBJECT) {
+				throw new InvalidTypeException(this.lineNo, ValueType.OBJ.toString(), property.type.toString());
+			}
 			object = property.objProperty;
 		} else {
 			object = environment.getObject(identifier.stringValue);
 		}
 		
-		//TODO: If distance is not a num, throw an exception
+		if (direction.type != ValueType.DIRECTION) {
```
akallem: I don't think this is necessary.  The grammar enforces that direction is a direction.  And it can never be null.  So there's nothing to check.  I suppose it doesn't hurt to have this, but perhaps a comment about it would be helpful for clarity.  An assertion would probably be even better, since it's an invariant we believe to be true and the user should never see.

src/INode.java
```diff
@@ -388,13 +407,17 @@ private ASTValue evalQueueSay(WordsEnvironment environment) {
...
+			if (property.type != WordsProperty.PropertyType.OBJECT) {
+				throw new InvalidTypeException(this.lineNo, ValueType.OBJ.toString(), property.type.toString());
+			}
 			object = property.objProperty;
 		} else {
 			object = environment.getObject(identifier.stringValue);
 		}
 		
-		//TODO: If message is not a string, throw an exception
+		if (message.type != ValueType.STRING) {
```
akallem, line 410: I realized my original comment was slightly wrong.  The reference manual allows either a string or a number (which is then coerced to a string).  So this should be changed to coerce a number to a string, or just open an issue about that point (which you can assign to me if you like).

src/INode.java
```diff
@@ -427,14 +450,12 @@ private ASTValue evalRemove(WordsEnvironment environment) {
...
-	private ASTValue evalRepeat(WordsEnvironment environment) {
+	private ASTValue evalRepeat(WordsEnvironment environment) throws WordsException {
 		ASTValue times = children.get(0).eval(environment);
 		AST statementList = children.get(1);
 		
 		if (times.type != ValueType.NUM) {
-			// TODO
-			// throw exception
-			return null;
+			throw new InvalidTypeException(this.lineNo, ValueType.NUM.toString(), times.type.toString());
```
akallem: My original code here was also inconsistent with the reference manual.  We allow a string if it can be coerced to a number.  You could either fix or open an issue for that and assign it to me if you like.

src/INode.java
```diff
@@ -461,7 +482,12 @@ private ASTValue evalSaysPredicate(WordsEnvironment environment) {
 
 	private ASTValue evalStatementList(WordsEnvironment environment) {
 		for (int i = 0; i < children.size(); i++) {
-			children.get(i).eval(environment);
+			try {
+				children.get(i).eval(environment);
+			} catch (WordsException e) {
+				System.err.println(e.toString());
+				System.err.print("> ");
```
akallem: Is the purpose of this to show the user the REPL prompt again?

src/ReferenceNotFoundException.java
```diff
@@ -0,0 +1,16 @@
+@SuppressWarnings("serial")
+public class ReferenceNotFoundException extends WordsException {
```
akallem: I don't think this is quite right.  It's allowed to retrieve a property an object where that property does not exist (per our discussions and the reference manual).  The result is simply the special value "nothing".  The error should occur only if the user dereferences "nothing".

src/InvalidTypeException.java
```diff
@@ -0,0 +1,16 @@
...
+	private String expectedType;
+	private String receivedType;
+	
+	public InvalidTypeException(int lineNo, String expected, String received) {
+		super(lineNo);
+		this.expectedType = expected;
+		this.receivedType = received;
+	}
+
+	public String toString() {
```
akallem: I think it's good practice to @Override the toString() method from Java.object so that one could print the exception without having to explicitly call toString() on it.  (This applies to all the exceptions, just putting the line comment here.)

src/WordsFunctionArgException.java
```diff
@@ -1,26 +1,20 @@
 
-public class WordsFunctionArgException extends Exception {
+@SuppressWarnings("serial")
+public class WordsFunctionArgException extends WordsException {
```
akallem: I don't quite understand how this one will work.  A function (custom action) can receive multiple parameters.  How do you map a single "expectedArg" to a single "receivedArg"?  Wouldn't it instead be a WordsFunctionMissingArgException, and the exception is simply the expectedArg that was missing?

src/WordsMove.java
```diff
@@ -59,7 +59,13 @@ public void doExecute(WordsObject object, WordsEnvironment environment) {
 	@Override
 	public LinkedList<WordsAction> doExpand(WordsObject object) {
 		if (distanceExpression != null) {
-			AST.ASTValue value = distanceExpression.eval(null);
+			AST.ASTValue value;
+			try {
+				value = distanceExpression.eval(null);
```
akallem: Putting null here (for the environment) won't work, because the expression being evaluated may include object property references, so it'll need access to the environment.  Which means the doExpand and expand methods will need the environment too.

src/WordsMove.java
```diff
@@ -59,7 +59,13 @@ public void doExecute(WordsObject object, WordsEnvironment environment) {
...
+			} catch (WordsException e) {
+				System.err.println(e.toString());
+				return null;
```
akallem: Will returning null here be a problem for actionQueue.addAll() in WordsObject.java?  It seems like we could get stuck in an infinite loop there.  I think the solution is in WordsObject.java to check if it's null, and if it is, return from executeNextAction entirely without attempting to do anything else.  You could pretty easily put that in or open a separate issue about it if you like.

src/WordsWait.java
```diff
@@ -34,7 +34,13 @@ public void doExecute(WordsObject object, WordsEnvironment environment) {
 	@Override
 	public LinkedList<WordsAction> doExpand(WordsObject object) {
 		if (lengthExpression != null) {
-			AST.ASTValue value = lengthExpression.eval(null);
+			AST.ASTValue value;
+			try {
+				value = lengthExpression.eval(null);
+			} catch (WordsException e) {
+				System.err.println(e.toString());
+				return null;
```
akallem: Same comment here as in doExpand() for WordsMove.

src/INode.java
alexthemark: Yes. It's going to print the error, which it would generally do on the same line as the current REPL prompt. The error printing is a little awkward since it comes up to a second after you typed things in.

I'm changing the second line to System.out, since it shouldn't be .err.

src/InvalidTypeException.java
alexthemark: I'll add in the override annotation to make it explicit (though Java overrides anyway without the annotation).

src/WordsFunctionArgException.java
alexthemark: I'm leaving that up to the programmer using the exception in this case. `receivedArg` and `expectedArg` are both String fields, so you can put in your list of arguments, and list of what you received. Think of it somewhat like the generic "usage" statement that most C programs have when they are run with the wrong number of arguments. You can say what you expected (the usage), and what you received (the type of each parameter).

src/INode.java
akallem: Add a comment too, since otherwise the print of > is pretty obscure.

src/WordsFunctionArgException.java
akallem: Gotcha.  Perhaps then it should be Args in all cases and not Arg, for clarity.

src/ReferenceNotFoundException.java
alexthemark: You're right. Changing this.

src/INode.java
alexthemark: Replacing with assertion.

src/INode.java
alexthemark: Fixed this by coercion.

src/INode.java
alexthemark, line 410: Fixed this by coercion.

alexthemark:
I'm thinking of separating our WordsExceptions into WordsProgramException, and WordsEnvironmentException. Currently, all the exceptions we have are WordsProgramExceptions, which contain line numbers and useful information for the programmer.

However, there are other times that we may want our environment to throw the exception (i.e. during object creation) when the environment doesn't know about how the program is being executed. I'm thinking that the environment can throw one of these exceptions, and iNode would likely catch it, and then repackage it into a WordsProgramException, which is more useful for the Words Programmer to see. I think this is also better than having the environment know what line number of the code we're on (I think that could very quickly get hairy). What are your thoughts on this?

I also haven't finished all updates to WordsMove and WordsWait that you noticed, @Aalk4308 . I bubbled down the environment so that the AST evaluation is useful, but there are some areas that require error handling. I have class for the next 5 hours, but if you have free time, feel free to fix these up so we can merge our error handling in asap. I'm not quite sure the semantics of where the errors in doExecute should be caught (perhaps in the frameloop)?

alexthemark:
The above commit handles what I was talking about. I modified the original idea slightly. A WordsProgramException just holds the environment's exception with a line number.

src/INode.java
```diff
@@ -320,9 +323,17 @@ private ASTValue evalParameterList(WordsEnvironment environment) {
...
 	}
 
-	private ASTValue evalPosition(WordsEnvironment environment) {
-		ASTValue row = children.get(0).eval(environment);
-		ASTValue col = children.get(1).eval(environment);
+	private ASTValue evalPosition(WordsEnvironment environment) throws WordsProgramException {
+		ASTValue row = children.get(0).eval(environment).getNumCoercedVal();
+		ASTValue col = children.get(1).eval(environment).getNumCoercedVal();
+		
+		if (row.type != ValueType.NUM) {
```
Aalk4308: You have a potential null pointer exception here (and in other similar places), because getNumCoercedVal() can return null.

src/AST.java
```diff
@@ -39,9 +39,48 @@ public ASTValue(WordsPosition p) {
...
+		 * Coerces the ASTValue into a num if it can be coerced.
+		 * 
+		 * @return self
+		 **/
+		public ASTValue getNumCoercedVal() {
+			if (type == ValueType.NUM) {
+				return this;
+			} else if (type == ValueType.STRING) {
+				  try  
+				  {
```
akallem: Technically this brace placement isn't consistent with our style guide (and other places like it).  :-)

Aalk4308:
OK, I'm OK for you to merge this.

alexthemark:
Great - I'll give it a few more hours in case anybody else has comments. Then we can merge, and everyone can branch off of this for their features this week.

akallem:
Perhaps a minor point, but I think "getNumCoercedVal" and "getStringCoercedVal" would be better named "coerceToNum" and "coerceToString" (or maybe even "tryCoerceToNum" and "tryCoerceToString").  (Or maybe even combining them into a single public method called "coerceTo" with the desired type as a parameter.)  I think of "get" methods as mainly property retrieval or simple lookups, not something that modifies the internal state of the object, which these do.

akallem:
I'm finding the two subclasses of exceptions rather confusing, especially with one being inside the other.  I understand what you're trying to do, but it feels a bit overengineered and very tricky.

I'm looking at WordsMove.java:63-67, and it's not obvious why that structure is the right structure compared to lines 69-71.

I'm also not finding getInnerException() to be a particularly instructive name.  Perhaps getEnvironmentException would be better?  But that's if we stick with this structure, which we should discuss.

This may be a case where a Javadoc comment at the top of the two main subclasses (as well as each specific exception subclass) would be really helpful, explaining the difference between the two main subclasses and how each is supposed to be used.

That said, I'm not sure what you think is potentially "hairy" about the other option, where the environment object knows what line number is currently being processed.  We just add one line at the top of eval() to set that field in the environment, and that's it: anywhere we could potentially throw an exception (which necessarily will have the environment, I think), we'll know the line number.  I'm not able to see the potentially adverse consequences you're concerned about.

I also renamed ReferenceException to WordsReferenceException, but I haven't yet for InvalidTypeException because I realized maybe there was a reason Words was not included in the front on those two but not the others?

alexthemark:
My meta-concern about having the environment know the line number being processed is that we're not encapsulating our data well at that point. The environment's job should be to keep track of the objects in the environment, and it really shouldn't know anything about the AST processing.

My more practical-level concern is that AST evaluation can happen during user input (when this method would be fine) and when dequeuing an action queue. When dequeueing an action queue, what would the current line number in the environment be? Would it be the last line number of the user commands that were just processed, or would the AST nodes from the action queue change the environment's line number (even though their line number could be a line number from some number of frames ago)? 

Lines 63-67 received a WordsProgramException, which contains an old AST line number, potentially from a number of frames ago. I didn't want that being printed, which is why I only threw the inner exception. I agree that some name refactoring/documentation can make these clearer.

akallem:
The "current line number" is not an ever-increasing number for each line that has been read in.  It's the line of the AST node that is currently being evaluated.  So, when dequeueing an action and evaluating some expression it contains, the line number is wherever that expression appeared in the source, which, conveniently, is the line number value on the AST node (and what would be the "current line number" in the environment if we set it at the top of eval()).

The assignment of line numbers to statements entered via REPL is a general and tricky question.  As it is now, the line numbers reset at each REPL prompt (and the user doesn't know what "line number" they're at).  So maybe line numbers aren't even the way to go, and we should reproduce the entire text of the line to make it clear to the user.

I don't know why you say an "old" AST line number in your example.  The line number on the AST object is the line number from the source on which that grammar expression appeared.  And it never changes.  So regardless of when that AST is processed, if there's a problem with it, it's line number is the right one to include in the error.

What if instead of the "current line number" being processed being a field in environment, it was a static field in the AST class?  It accomplishes the same thing I was suggesting, but I agree it's a bit better encapsulation, because the environment itself doesn't know the line number directly.

akallem:
Looks good, let's merge it.

PR #34, opened 2015-04-02
Boolean/Relational Operators and If Statements

akallem:
This branch contains implementations of the boolean and relational operators as well as the If statement.  It also contains JUnit tests for all of the implementations (except If, which I suggest we test in actual test programs instead of via unit tests).

For consistency, I'd suggest that the arithmetic operators use a similar format and structure (with lhs and rhs variables).

Addresses #20, #21, #25, #26, #27, #28, #29, #31.


src/INode.java
```diff
@@ -229,9 +251,63 @@ private ASTValue evalDefineProperty(WordsEnvironment environment) {
...
+				if (rhs.type == ValueType.STRING)
+					lhs = lhs.tryCoerceTo(ValueType.STRING);
+			}
+		}
+		
+		// lhs and rhs are now the same type
+		assert lhs.type == rhs.type : "lhs has type " + lhs.type.toString() + " and rhs has type " + rhs.type.toString();
+		switch (lhs.type) {
+			case NUM:
+				return new ASTValue(lhs.numValue == rhs.numValue);
```
alexthemark, line 305: We should probably include some error tolerance since we're using doubles.

src/INode.java
```diff
@@ -244,14 +320,38 @@ private ASTValue evalExponentiate(WordsEnvironment environment) {
...
+	private ASTValue evalGEQ(WordsEnvironment environment) throws WordsRuntimeException {
+		ASTValue lhs = children.get(0).eval(environment);
+		ASTValue rhs = children.get(1).eval(environment);
+		
+		checkRelOpArgTypes(lhs, rhs);
+		
+		// lhs and rhs are now the same type
+		switch (lhs.type) {
+			case NUM:
+				return new ASTValue(lhs.numValue >= rhs.numValue);
```
alexthemark, line 332: Once again, some error tolerance for the equals part of geq. (Same comment applies to leq).

alexthemark:
Now that I also have the template for JUnit testing up, it would be great if you could automate your test cases.

src/INode.java
akallem, line 305: Any idea how other languages handle this?  Also, is it actually necessary?  However Java is doing it, aren't we essentially just piggy-backing off of it?

akallem:
Agreed, I'll definitely do that!

src/INode.java
alexthemark, line 305: I think that, for a beginner language without integer representation, this is a good feature to have. It's an area where I would say Java is not intuitive, and we can choose to be.

akallem, line 305: OK.  As language guru, you can propose what you think our semantics should be, and then I'll implement it.  :-)

PR #35, opened 2015-04-03
Add template for test cases, first test cases

### LINKED ISSUE #18: Create JUnit test framework
Create 2 examples -- one of a mutator and one of an inode that returns values.

alexthemark:
Closes #18 

This PR contains a number of tests on existing code, all using the JUnit framework. I'm additionally going to create issues for INodes that are currently not tested (I have obligations for the rest of the weekend so I'm not going to be able to get to them -- @Aalk4308 , since you wrote most of the things that need testing, would you like to ensure that they actually work?)

The TestEnvironment class provides some documentation about how to write JUnit tests, but I'd recommend reading over http://www.vogella.com/tutorials/JUnit/article.html#junit_intro for a short introduction. For future PRs, including this week's features, everybody should have comprehensive testing.


src/WordsPosition.java
```diff
@@ -9,12 +9,18 @@ public WordsPosition(int x, int y) {
...
 	}
 	
 	@Override 
 	public String toString() {
 		return Integer.toString(x) + " , " + Integer.toString(y);
 	}
+	
+	@Override
+	public boolean equals(Object other) {
+		WordsPosition otherPos = (WordsPosition) other;
```
akallem, line 26: Technically, before casting, I think you should check if 'other' is an instance of WordsPosition.  Otherwise, the cast would result in an error.  We could even use EqualsBuilder, though I don't care too much about that part.  See this post: http://stackoverflow.com/questions/27581/what-issues-should-be-considered-when-overriding-equals-and-hashcode-in-java

test/TestINode.java
```diff
@@ -0,0 +1,38 @@
...
+public class TestINode {
+	
+	WordsEnvironment environment = new WordsEnvironment();
+
+	/*********************
+	 * EvalPosition 
+	 ********************/
+	AST numLeaf1 = new LNode(AST.ASTType.NUM, 0, 0.0);
+	AST numLeaf2 = new LNode(AST.ASTType.NUM, 0, 2.1);
+	AST stringLeafCoercable = new LNode(AST.ASTType.STRING, 0, "2.1");
```
akallem: It seems like you're suggesting we would pre-create a lot of "ready-to-use" AST nodes that can be used in different combinations in each test case.  But since the coercion method modifies the AST node itself, won't the coercion happen just one time (by whichever test is run first), and later tests that use this node will not really be testing whether they have incorporated coercion correctly?

src/WordsPosition.java
alexthemark, line 26: Fixing.

test/TestINode.java
alexthemark: JUnit has a fun feature where all instance variables are actually reinitialized between each test case. I've modified the TestEnvironment file to better document and show this. 

Though you have a fair point that it's not really necessary to test coercion separately in each INode test case, as long as we have extensive testing of coercion. I'm going to write up these test cases real fast.

alexthemark:
I've added tests for coercion, so that inode tests don't have to repeat it (though I've left the slightly redundant test case in here anyway. No harm (except run time, and these are like .0002s) in leaving a working test). To construct ASTValue outside the AST class, I had to move it to its own top level class. Separately, I think that it's grown large enough to deserve this.

akallem:
Looks good overall.  The only thing I'm finding is that as I'm writing test cases for the INode class, I think it would be easier if each test case was fully self-contained (aside from the environment).  On the one hand, having some common variables (numLeaf1, stringLeafCoercable) shared across tests might reduce duplication, but on the other hand it makes it harder for each of us to right our own tests.  I'm going to make a slight edit that I think will make this easier going forward.

akallem:
One other thought: For TestINode, I think we should focus on unit tests for the "simple" methods like logical operators, arithmetic, etc.  For the more complex nodes (iteration, listeners, etc.), I don't think it will be easy to write unit tests, since their effects are complex.  Instead, we should test those constructs in actual test programs.  Thoughts?

alexthemark:
I'm in complete disagreement. If something is difficult to test, that's something to think about or to fix. It's common practice that a test suite encompasses your entire code base, and that works on codebases far larger and more complicated than ours. I'm going to upload changes in about half an hour that should test iteration, and provide easier access into the environment for testing these larger cases (and hopefully help your skepticism towards test suites). 

My biggest problem with logging is that we're not testing what's actually happening in the code or in the environment -- we're testing a string representation of that. That means that, if we change our logging method, we need to rebuild all logs. Or if we change a string (i.e. how errors are printed), we need to rebuild all logs. Any level of log automation that we build is all something that needs to be built -- and tested. I also don't know if I can trust the logs -- any bug in it could be these representation issues or an issue with how we're logging, or an issue with the actual code. They also just prove that our program ran the same way it ran before, instead of actually running correctly. We have JUnit, which can be used to automate all of this away by actually looking at the objects.

alexthemark:
We can decide in our meeting tomorrow how to test larger pieces, but commit 7f109a1 should provide a pretty good format. 

I'd like to get this merged into master very soon so that everyone can integrate it into their feature branches.

zwdnyc:
I agree that we should definitely test what's actually happening in the code or in the environment. This is what JUnit is used for. But I think it would be even better if we perform tests at different levels as well (http://en.wikipedia.org/wiki/Software_testing#Testing_levels). To me, the logging method is a mechanism for implementing system testing. Because our output is a visual representation of things updated at some time interval, it might be slow to run such tests for every Words test program we written. Instead, we dump such visual representations as logs, which is faster to run and easier to compare.

We can first run the test programs and check if they work correctly. If there is no problem, we build the logs. Then, for any subsequent changes in our implementation, we can just run the automated log comparing tests to see if our changes do not introduce any side effects in the final log output (as a proxy of visual output), without actually observing and comparing the visual output. I guess this should be easier for us to ensure many test Words program work in our expected ways.

zwdnyc:
I have a question though. Is it possible that the output of some Words program is non-deterministic? If so, then the logs cannot be used for testing them.

alexthemark:
Yes. Any place where we say "Make [object] move anywhere."

But we can just avoid those in logging tests.

akallem:
I'm OK for this to be merged.

akallem:
Logs could still be used in that case, as long as the Words interpreter is
launched with a "test" flag (which it would need to be, anyway, to suppress
the GUI and increase the frame rate), and the test flag causes the random
number generator to be seeded with a fixed value.

PR #36, opened 2015-04-03
Fix Action Queue bug

### LINKED ISSUE #33: Action Queue after Exception
In WordsObject.java:executeNextAction(), I don't think it's necessary to pop a second time before throwing the exception up.  The problematic action has already been popped, and nothing has been added back to the queue (because the exception was thrown).

alexthemark:
Closes #33 

I hadn't noticed the pop on line 100 when I originally wrote the code... only the peek on line 99. The second pop is definitely wrong. I think the first person who sees this can probably merge it into master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ hs_err_pid*
 *.externalToolBuilders
 *.settings
 *.project
+/bin/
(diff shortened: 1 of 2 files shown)


akallem:
OK, this fix looks good.

PR #39, opened 2015-04-05
Eval ref list eval assign

mzbenami:
Hey guys!
Pull requests are fun!

This is my attempt at the implementations for evalReferenceList() and evalAssign().

I wasn't sure if it was my possibility to throw exceptions like WordsReferenceException or if that would be initiated by the runtime, so I just bubble all WordsExceptions up.

I hope I understood the semantics of NOTHING vs. null correctly, please let me know if I missed that.

Thanks,
Michael


src/INode.java
```diff
@@ -189,9 +189,29 @@ private ASTValue evalAnd(WordsEnvironment environment) {
...
+				wordsProp = new WordsProperty(propertyASTValue.stringValue);
+				break;
+			case OBJ:
+				wordsProp = new WordsProperty(propertyASTValue.objValue);
+				break;
+			default:
+				wordsProp = new WordsProperty(WordsProperty.PropertyType.NOTHING);
+		}
+	
+		obj.setProperty(propertyName, wordsProp);
```
alexthemark: Can you fix set property so it throws the exception it's supposed to (just took a look in WordsObject.setProperty and it has a to do. Figure that this branch is a good place to fix that)?

src/INode.java
```diff
@@ -189,9 +189,29 @@ private ASTValue evalAnd(WordsEnvironment environment) {
 		throw new AssertionError("Not yet implemented");
 	}
 
-	private ASTValue evalAssign(WordsEnvironment environment) {
-		// TODO
-		throw new AssertionError("Not yet implemented");	
+	private ASTValue evalAssign(WordsEnvironment environment) throws WordsRuntimeException {
+		WordsObject obj = children.get(0).eval(environment).objValue;
```
alexthemark: This isn't ever making a call to the environment to get an object. Does this work? I think that the 0th child is actually storing the object name.

src/INode.java
```diff
@@ -431,10 +451,24 @@ private ASTValue evalQueueWait(WordsEnvironment environment) {
 		throw new AssertionError("Not yet implemented");
 	}
 
-	private ASTValue evalReferenceList(WordsEnvironment environment) {
-		// TODO
-		//throw new AssertionError("Not yet implemented");
-		return new ASTValue(ValueType.NOTHING);
+	private ASTValue evalReferenceList(WordsEnvironment environment) throws WordsRuntimeException {
+		String firstObjRefName = children.get(0).eval(environment).stringValue;
+		WordsObject currentObj = environment.getObject(firstObjRefName);
```
alexthemark: Right now, getObject can return null, and you'd need to throw an exception from this code in case currentObj is null. I'd be fine with also modifying get object to throw the exception, but we do need to check.

alexthemark:
Can you also add unit testing for assignment and retrieval? You should be able to test a single reference, a reference chain, as well as any exceptions that occur. You should probably merge in the testing branch (which also contains documentation on how to test).

src/INode.java
mzbenami: yes, i think you're right. will fix.

akallem:
Also, there should be at least some .words program that actually utilizes these features to provide a baseline check that they actually work (in addition to the JUnit tests).

src/INode.java
akallem: Actually, per the grammar, the first (index zero) child is a REFERENCE_LIST node, and the evaluation of this node should indeed return an object (not a string).  So I think the existing code is correct.

src/INode.java
akallem: Let's make that an issue and handle it separately from this branch.  I've just added it as an issue.

src/INode.java
```diff
@@ -431,10 +451,24 @@ private ASTValue evalQueueWait(WordsEnvironment environment) {
...
+		
+		for (int i = 1; i < children.size(); i++) {
+			String propertyRefName = children.get(i).eval(environment).stringValue;
+			WordsProperty prop = currentObj.getProperty(propertyRefName);
+			if (prop.type != WordsProperty.PropertyType.OBJECT) {
+				return new ASTValue(ValueType.NOTHING);
```
akallem: I think this should be an exception, not NOTHING.  A reference list is a chain of objects, the first provided by name, the others provided as property names.  At every step (including the last), it must be an object.  Otherwise, the user would be trying to dereference a non-object, which is an error on their part.

mzbenami:
Almost done guys. Just need to write 2 more JUnit tests.

mzbenami:
hey guys, i am happy with this code, and the system level test at system_test/PropertiesTest.words which tests:
- property assignment to object
- reference list evaluation
- evaluation of value-expressions when given by reference reference_list identifier syntax
- new way of handling "object not found" and "object already exists" errors.

However, I am struggling with the one JUnit test I wrote, TestINodeAssign.java. It fails and I can't figure out why. Is there a place I could look or a method I can try to debug the JUnit tests?

alexthemark:
Hey -- I merged in master so that we're able to merge. I also fixed the JUnit error. I left a comment in your last commit showing what caused it. 

We also need a PropertiesTest.words.log file committed so that we have a comparison for future system tests.

test/words/test/TestINodeAssign.java
```diff
@@ -0,0 +1,36 @@
...
+		INodeAssign objAssign = new INodeAssign(alexRefList, new LNodeIdentifier("friend"), new INodeReferenceList(), bobIdentifier);
+		INodeAssign nothingAssign = new INodeAssign(alexRefList, new LNodeIdentifier("myNothing"), nothingLeaf);
+		
+		
+		numAssign.eval(environment);
+		stringAssign.eval(environment);
+		objAssign.eval(environment);
+		nothingAssign.eval(environment);
+		
+		assertEquals("Number assignment successful", alexObject.getProperty("height").numProperty, 5.0, .0001);
```
alexthemark: It's often best practice to separate out these tests -- create four different tests, one for String, one for number, one for object, and one for nothing. That way, if one of these tests fail, we only see that specific functionality has failed. 

Line numbers will sort of do this, but will depend on what fails first. I.e. if number assignment failed, then we have no idea if string assignment works, because the program failed before it reached there. This information is often useful for the people who break your tests in future updates.

test/words/test/TestINodeAssign.java
```diff
@@ -0,0 +1,36 @@
...
+import words.ast.*;
+import words.exceptions.*;
+import words.environment.*;
+
+import org.junit.Test;
+
+public class TestINodeAssign extends TestINode {
+	
+	@Test
+	public void assignProperty() throws WordsRuntimeException {
```
alexthemark: We should also have a test that this WordsRunTimeException is thrown when it should be thrown.

mzbenami:
Ok guys, I think we are close. This might be it. To recap:
- INodeRetrieveProperty and JUnit tests
- INodeAssign and JUnit tests
- INodeReferenceList and JUnit tests
- summary system test PropertiesTest.words

JUnit tests include all inputs I can think of and all possible exceptions.
My only issue now is that I put the test folder on my build path in Eclipse so I can run JUnit right from eclipse, and now 'ant test' fails all tests on the command line, including all of the originals. If I remove from build path, 'ant test' succeeds. I suppose this is a minor detail and can be worked out later.

mzbenami:
Oh, I forgot:
- log PropertiesTest.words.log

akallem:
OK, so should we review it now?

mzbenami:
Yes please

src/words/environment/WordsEnvironment.java
```diff
@@ -54,31 +55,40 @@ public WordsClass getClass(String className) {
...
-		
-		WordsClass wordsClass = getClass(className);
+		} catch (WordsObjectNotFoundException e){
 		
-		if (wordsClass == null) {
-			throw new WordsClassNotFoundException(className);
+			WordsClass wordsClass = getClass(className);
+			
+			if (wordsClass == null) {
+				throw new WordsClassNotFoundException(className);
```
alexthemark: I think that WordsEnvironment.getClass should probably throw the class not found exception, instead of checking for null here. That's the pattern we established with .getObject

test/words/test/TestINodeRetrieveProperty.java
```diff
@@ -0,0 +1,86 @@
...
+		WordsObject alexObject = environment.createObject("Alex", "thing", new WordsPosition(0,0));
+		LNodeReference alexLNodeRef = new LNodeReference("Alex's");
+		INodeReferenceList alexRefList = new INodeReferenceList(alexLNodeRef);
+		
+		alexObject.setProperty("occupation", new WordsProperty("model"));
+		
+		INodeRetrieveProperty retrieveAlexOccupation = new INodeRetrieveProperty(alexRefList, new LNodeIdentifier("occupation"));
+		ASTValue alexOccupation = retrieveAlexOccupation.eval(environment);
+		
+		assertEquals(alexOccupation.stringValue, "model");
```
alexthemark, line 64: :fu:

alexthemark:
Ant test is working for me when I've pulled your branch. Maybe try running an ant clean, and then ant test? Residual files can sometimes cause an issue.

src/words/ast/INodeAssign.java
```diff
@@ -8,9 +10,36 @@ public INodeAssign(Object... children) {
...
 
+	/**
+	 * Always returns null. Throws WordsRuntimeException for called methods, including WordsObjectNotFoundException.
+	 * Side effect of assigning a property to an object.
+	 */
 	@Override
 	public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException {
-		// TODO
-		throw new AssertionError("Not yet implemented");
+		WordsObject obj = children.get(0).eval(environment).objValue;
```
akallem, line 19: We could put in an assert here that the obj is not null.  In principle. the child (a reference list) could have been NOTHING, but the grammar right now assures that that's not the case.  If we change the grammar, we might miss that.  I guess we'd catch that regardless via a null pointer exception, but an assertion with a comment might be clearer.

src/words/exceptions/WordsReferenceException.java
```diff
@@ -2,10 +2,17 @@
...
 public class WordsReferenceException extends WordsRuntimeException {
 	
+	private String message;
+	
+	public WordsReferenceException(String message) {
+		this.message = message;
+	}
+	
 	public WordsReferenceException() {
+		this.message = "attempted to dereference nothing";
```
akallem: I think this isn't quite the right description for how this exception is being used.  This exception is being used any time we try to dereference something that is not an object.  That includes NOTHING, but it also includes anything else too (number, string).

I see that there is an alternative constructor with a message directly applied.  That's different from how we've done every other exception.  I'd suggest it instead have two fields: the type and the identifier.  Then the message would be generated as "Attempted to dereference [identifier] which has invalid type [type]" or something like that.

src/words/ast/INodeRetrieveProperty.java
```diff
@@ -9,8 +9,54 @@ public INodeRetrieveProperty(Object... children) {
...
-		throw new AssertionError("Not yet implemented");
+		ASTValue refList = children.get(0).eval(environment);
+		if (refList.type == ASTValue.ValueType.NOTHING) {
+			ASTValue id = children.get(1).eval(environment);
+			
+			try {
+				WordsObject obj = environment.getObject(id.stringValue);
+				return new ASTValue(obj);
+			} catch (WordsObjectNotFoundException e) {
+				throw new WordsReferenceException("Value expression is not an object");
```
akallem: Similar to my comment in WordsReferenceException, we should avoid passing a message string itself when creating the exception, because then we repeat that message multiple times (like below in this file).

src/words/ast/INodeRetrieveProperty.java
```diff
@@ -9,8 +9,54 @@ public INodeRetrieveProperty(Object... children) {
...
+			}
+		}
+		
+		WordsObject obj = refList.objValue;
+		if (obj == null) {
```
akallem, line 30: Is this condition possible?  If INodeReferenceList doesn't throw an exception, then it only returns NOTHING (which is handled above) or an object (which is here).  So I don't think obj could ever be null.  If you agree, then let's make this an assertion, which is more appropriate.

src/words/ast/INodeRetrieveProperty.java
```diff
@@ -9,8 +9,54 @@ public INodeRetrieveProperty(Object... children) {
...
+			throw new WordsReferenceException("Value expression is not an object");
+		}
+		
+		ASTValue id = children.get(1).eval(environment);
+		String propName = id.stringValue;
+		
+		WordsProperty wordsProp = obj.getProperty(propName);
```
akallem, line 37: Since this code is basically the same as in INodeAssign, I think it would be better to make it a  method of the class.  For example, the ASTValue class could have a method called "toWordsProperty()" or something like that which does this conversion.

test/words/test/TestINodeRetrieveProperty.java
akallem, line 64: Hahaha, that is fantastic.

test/words/test/TestINodeReferenceList.java
```diff
@@ -0,0 +1,67 @@
...
+		WordsObject alexObject = environment.createObject("Alex", "thing", new WordsPosition(0, 0));
+		LNodeReference alexRef = new LNodeReference("Alex's");
+		INodeReferenceList alexRefList = new INodeReferenceList(alexRef);
+		
+		ASTValue alexValue = alexRefList.eval(environment);
+		assertEquals(alexValue.objValue, alexObject);
+	}
+	
+	@Test
+	public void testSizeTwo() throws WordsRuntimeException {
```
akallem, line 32: For thoroughness, could we make this a test of three (or even four) instead of just two?  I'd feel more confident that that will capture the general case, whereas effecitvely only captures one iteration of the reference loop and therefore doesn't test the iteration.

src/words/environment/WordsEnvironment.java
mzbenami: i agree, fixed it.

src/words/ast/INodeAssign.java
mzbenami, line 19: Ah, I see what you're saying. You're right I depended on the grammar enforcing that. Since at least one reference must be in the list for that production, there's either an object or an exception will be thrown. I'll throw the AssertionError. Am I correct in understanding our use of the AssertionError as something that should never happen under the current way we have things set up?

src/words/ast/INodeRetrieveProperty.java
mzbenami: Yes, I changed this to just throw the original WordsObjectNotFoundException.

src/words/ast/INodeRetrieveProperty.java
mzbenami, line 30: You're right. Throwing an AssertionError.

src/words/ast/INodeRetrieveProperty.java
mzbenami, line 37: Which ASTValue are you talking about? The one representing the object or the one representing the property? It only makes sense to me if you're talking about the one representing the property. In which case do you mean the switch statement that follows the code highlighted here?

test/words/test/TestINodeReferenceList.java
mzbenami, line 32: Yup made it four.

mzbenami:
I implemented all the suggestions besides Andrew's of the ASTValue method for property conversion.

src/words/ast/INodeAssign.java
akallem, line 19: Yes that's right, we're using Assertions as a defensive coding practice for ourselves, not something the user should ever see.  if we trigger an AssertionError during our development, it means our assumptions about our code were wrong, or we've changed things such that those assumptions are no longer true, either of which will be good to have caught.

src/words/ast/INodeRetrieveProperty.java
akallem, line 37: Ah, my mistake, this code is symmetric with the code in INodeAssign but not the same.  This is WordsProperty to ASTValue, whereas INodeAssign is ASTValue to WordsProperty.  So they're fine as is.  (Although if down the road we find the need to do either of those again, then we could factor it out into a separate method.)

akallem:
Looks great, I'm good with this.  (Responded with a line note that the suggestion I made about ASTValue method was wrong, so it's good as is.)

akallem:
@ZhangWangda and @jtyoung1292 can you also review?

youngthejames:
I'm good with this too.

src/words/ast/INodeRetrieveProperty.java
```diff
@@ -9,8 +9,51 @@ public INodeRetrieveProperty(Object... children) {
...
+			ASTValue id = children.get(1).eval(environment);
+			
+			WordsObject obj = environment.getObject(id.stringValue);
+			return new ASTValue(obj);
+
+		}
+		
+		WordsObject obj = refList.objValue;
+		if (obj == null) {
+			new AssertionError("obj shouldn't be null");
```
akallem: Need to throw the assertion here, not just create it.

src/words/ast/INodeRetrieveProperty.java
```diff
@@ -9,8 +9,51 @@ public INodeRetrieveProperty(Object... children) {
...
+				astValue = new ASTValue(wordsProp.numProperty);
+				break;
+			case OBJECT:
+				astValue = new ASTValue(wordsProp.objProperty);
+				break;
+			case NOTHING:
+				astValue = new ASTValue(ASTValue.ValueType.NOTHING);
+				break;
+			default:
+				new AssertionError("Shouldn't get here in INodeRetrieveProperty");
```
akallem: Need to throw the assertion here, not just create it.

mzbenami:
gosh i feel silly. this thing is dying to be merged...one day. in the meantime, i'm really impressed with the testing environment we've got set up. now that i've seen i don't know how i functioned without it.

PR #41, opened 2015-04-06
Arithmetic operators

### LINKED ISSUE #23: Math
evalNegate
evalAdd
evalSubtract
evalMultiply
evalDivide
evalExponentiate

youngthejames:
Closes #23


src/INode.java
```diff
@@ -229,6 +244,17 @@ private ASTValue evalDefineProperty(WordsEnvironment environment) {
 		throw new AssertionError("Not yet implemented");	
 	}
 
+	private ASTValue evalDivide(WordsEnvironment environment) throws WordsRuntimeException {
+		ASTValue lhs = children.get(0).eval(environment).tryCoerceTo(AST.ValueType.NUM);
+		ASTValue rhs = children.get(1).eval(environment).tryCoerceTo(AST.ValueType.NUM);
+
+		if((lhs.type != AST.ValueType.NUM) || (rhs.type != AST.ValueType.NUM)) {
+			throw new AssertionError("Operation arguments must both be ValueType NUM");
```
akallem: We should only use Assertions for conditions that we expect to always be true because our implementation or the semantics of the language guarantee it.  Assertions are a way for us to ensure we haven't made an error.  In this case, however, a user could (erroneously) try to divide things that aren't numbers.  In this case, it's not an AssertionError exception, it's a WordsRuntimeException that we will eventually catch and display to the user.  You can pick from the among the existing exceptions (OperatorTypeMismatch might be a good one) or create a new one.  This comment applies not just to this method but to all of the eval arithmetic methods.

test/arithmetic.words
```diff
@@ -0,0 +1,17 @@
...
+
+Make Fred say 3*4.
+
+Make Fred say 4/3.
+
+Make Fred say 4^3.
+
+Make Fred say 4+3.
+
+Make Fred say 6+"7". // 6.0000007
```
akallem: Why would this result in 6.0000007?  Shouldn't it result in 13?

src/INode.java
```diff
@@ -174,9 +175,23 @@ public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException
...
+	private ASTValue evalAdd(WordsEnvironment environment) throws WordsRuntimeException {
+		ASTValue lhs = children.get(0).eval(environment);
+		ASTValue rhs = children.get(1).eval(environment);
+		
+		if(((lhs.type == AST.ValueType.STRING) && (rhs.type == AST.ValueType.NUM)) || 
+				((lhs.type == AST.ValueType.NUM) && (rhs.type == AST.ValueType.STRING)) ||
+				((lhs.type == AST.ValueType.STRING) && (rhs.type == AST.ValueType.STRING))) {
+			lhs.tryCoerceTo(AST.ValueType.STRING);
+			rhs.tryCoerceTo(AST.ValueType.STRING);
+			return new ASTValue(lhs.stringValue+rhs.stringValue);
```
Aalk4308: These semantics are actually different than what is described in the Reference Manual.  Here, a number is always coerced to a string if the other operand is a string, whereas the Reference Manual first says to try coercing the string to a number and use addition if it succeeds.

src/INode.java
```diff
@@ -229,6 +244,17 @@ private ASTValue evalDefineProperty(WordsEnvironment environment) {
 		throw new AssertionError("Not yet implemented");	
 	}
 
+	private ASTValue evalDivide(WordsEnvironment environment) throws WordsRuntimeException {
+		ASTValue lhs = children.get(0).eval(environment).tryCoerceTo(AST.ValueType.NUM);
+		ASTValue rhs = children.get(1).eval(environment).tryCoerceTo(AST.ValueType.NUM);
+
+		if((lhs.type != AST.ValueType.NUM) || (rhs.type != AST.ValueType.NUM)) {
```
akallem: We should try to adhere to our style guide as best we can, which means there should be a space between the "if" and the "(".

src/INode.java
```diff
@@ -174,9 +175,23 @@ public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException
...
+		ASTValue rhs = children.get(1).eval(environment);
+		
+		if(((lhs.type == AST.ValueType.STRING) && (rhs.type == AST.ValueType.NUM)) || 
+				((lhs.type == AST.ValueType.NUM) && (rhs.type == AST.ValueType.STRING)) ||
+				((lhs.type == AST.ValueType.STRING) && (rhs.type == AST.ValueType.STRING))) {
+			lhs.tryCoerceTo(AST.ValueType.STRING);
+			rhs.tryCoerceTo(AST.ValueType.STRING);
+			return new ASTValue(lhs.stringValue+rhs.stringValue);
+		}
+		else if((lhs.type == AST.ValueType.NUM) && (rhs.type == AST.ValueType.NUM)) {
```
akallem: We should follow our style guide, which means the else should be on the same line as the brace, i.e., 
if {
   ...
} else if {
   ...
} else {
   ...
}

src/INode.java
```diff
@@ -229,6 +242,17 @@ private ASTValue evalDefineProperty(WordsEnvironment environment) {
 		throw new AssertionError("Not yet implemented");	
 	}
 
+	private ASTValue evalDivide(WordsEnvironment environment) throws WordsRuntimeException {
+		ASTValue lhs = children.get(0).eval(environment).tryCoerceTo(AST.ValueType.NUM);
+		ASTValue rhs = children.get(1).eval(environment).tryCoerceTo(AST.ValueType.NUM);
+
+		if ((lhs.type != AST.ValueType.NUM) || (rhs.type != AST.ValueType.NUM)) {
+			throw new WordsArithmeticException(lhs.type.toString(), rhs.type.toString());
```
Aalk4308: It will be confusing to the user if the lhs is a number and rhs is a string, (so the condition is true), but the exception thrown and ultimately displayed to the user says number (since lhs is a number).  Perhaps it's better to check each operand independently for lhs and rhs, and throw the exception with lhs or rhs type, so that the type shown to the user is actually the one that caused the problem.

src/INode.java
```diff
@@ -174,9 +175,21 @@ public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException
...
+		if (((lhs.type == AST.ValueType.STRING) && (rhs.type == AST.ValueType.NUM)) || 
+				((lhs.type == AST.ValueType.NUM) && (rhs.type == AST.ValueType.STRING)) ||
+				((lhs.type == AST.ValueType.STRING) && (rhs.type == AST.ValueType.STRING))) {
+			lhs.tryCoerceTo(AST.ValueType.STRING);
+			rhs.tryCoerceTo(AST.ValueType.STRING);
+			return new ASTValue(lhs.stringValue+rhs.stringValue);
+		} else if ((lhs.type == AST.ValueType.NUM) && (rhs.type == AST.ValueType.NUM)) {
+			return new ASTValue(lhs.numValue+rhs.numValue); 
+		} else {
+			throw new WordsArithmeticException(lhs.type.toString(), rhs.type.toString());
```
akallem: I think you may have forgotten to "git add" WordsArithmeticException.java, since it's not included on this branch.

alexthemark:
Can you add JUnit test cases?

youngthejames:
working on it.

youngthejames:
Added tests and everything builds. I think it's ready to merge.

akallem:
Looks great, I'm good for this to be merged.

src/words/exceptions/WordsArithmeticException.java
```diff
@@ -0,0 +1,17 @@
...
+	private String type2;
+	
+	public WordsArithmeticException(String type1, String type2) {
+		this.type1 = type1;
+		this.type2 = type2;
+	}
+
+	@Override
+	public String toString() {
+		return String.format("Invalid arithmetic operation on %s and %s", type1, type2);
```
alexthemark: Should have a period on the end of the string for consistency.

test/words/test/TestINodeNegate.java
```diff
@@ -0,0 +1,25 @@
...
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+import words.ast.*;
+import words.exceptions.*;
+
+public class TestINodeNegate extends TestINode {
+	@Test
+	public void twoNumbersNegateCorrectly() throws WordsRuntimeException {
```
alexthemark: I think this test should be named for one number negating correctly, not two.

youngthejames:
Sorry I fixed it.
Error made in haste.

mzbenami:
Looks good to me.

akallem:
Cool.  @ZhangWangda can you review too?

PR #43, opened 2015-04-07
Refactor and cleanup

akallem:
This branch makes each type of LNode and INode into a separate subclass, which will make it easier for us to work concurrently in different files.  Likewise, the JUnit tests for INode have been separated into distinct files.  It also includes some bug fixes and some minor documentation and style fixes.

Last, I modified the constructor of AST nodes so that they only ever take the value (leaf node) or children (internal node).  The line number is a public field and can be set after the node is created.  You can see this being done in the grammar (Words.y).  The reason for this is so that the JUnit tests, which create lots of nodes, don't have to repeatedly include a bogus line number (e.g., 0) in their constructor, which wasn't particularly pretty.


zwdnyc:
I just had a minor modification to the test scripts. Everything else looks good to me.

alexthemark:
This is a solid refactor, and I'm really happy that we decided to do this. Thanks for putting in the effort to make our code base better for the future. I glanced over the diff, but didn't do a full read (too difficult/long when most of it is just code copy pasted into new files). 

I will say that I disagree with changing a constructor to make tests a little cleaner (especially when it causes you to always set a public field immediately after constructing). Tests often have bogus data and data that's untrue, and that doesn't matter (as long as that data doesn't matter). However, I don't care enough about this to justify a) fighting over it or b) going through and changing all the constructors of nodes. 

I'm good with a merge to master.

akallem:
Great!  FYI, I think there may be legitimate cases where we want to create AST nodes with no line numbers in the normal code base, i.e., for the REPL interface.

Michael and James, please give it a review ASAP so that we can merge.

youngthejames:
Everything looks good to me.

mzbenami:
I approve of this. Nothing jumped out as a mistake or bug.

akallem:
Great, I will merge now.

PR #44, opened 2015-04-07
Automated system test

akallem:
This branch takes Wangda's system testing code and integrates it with our ant build system.  Now, when you run "ant test", it will run both the JUnit tests and the system integration tests.  It will report an error if any JUnit test fails and will abort with an error as soon as the first system integration test fails (because it was simpler to implement this way).

Previously, the no-GUI option would still have the programs run with a small delay between frames.  I now have them run with zero delay between frames so they run as fast as possible.

Also, previously the system integration testing set a 20 frame cap.  Some sort of cap may be necessary for long programs, but I've raised this to 500 (which should be more than sufficient and still runs extremely quickly), and I've ensured it only takes effect in test mode and not normal use.

NOTE: This branch is based off of the refactor-and-cleanup branch, so we'll probably merge that first, then merge this.


zwdnyc:
Thanks a lot, Andrew, for completing the tests. I'm glad to merge this.

alexthemark:
Looks good and runs on my computer. Andrew, have you checked that it runs on Windows/your environment? Text files created on windows have different line endings ("\r\n") than Unix ("\n"), and I'm worried about diffs failing due to that.

akallem:
Great.  Yes, I took care of the \r\n vs. \n in my text editor.  I'm on Windows and the .bat scripts run fine for me.  And I used Cygwin to test the Unix .sh scripts, which worked there, and I'm glad to see they worked on yours too.  So everything should be all set.

PR #45, opened 2015-04-07
Refactor packages

alexthemark:
Branched off of Andrew's refactor branch. 

The number of files that we have under the default package has become a bit unwieldy, but there are pretty natural groupings that we can put into packages. AST related things go into words.ast; exceptions go into words.exception; and anything to do with how we represent our environment goes into words.environment.

I just noticed that I'm having some residual effects on running the test suite, so give me a few hours to figure these out.


akallem:
OK, will you post a comment again when you've figured out those issues so we know when to take another look?  Alternatively, I guess you could close the pull request and reopen it when you're ready, right?

PR #46, opened 2015-04-07
Refactor packages

alexthemark:
Organize the code base into various packages, since it was very large. 

Test cases work now. There are no real changes to code except for the now-necessary import statements. Since we're no longer using the default package for everything, it is sometimes necessary to import various parts of our code (i.e. exceptions), but this seems normal for most Java code bases.


akallem:
This looks great, I'm good with this.

alexthemark:
I think we should probably merge mine and Wangda's branches tonight so that Michael and James can incorporate them for their Friday update. @mzbenami @jtyoung1292 it would be great if you had time to look over these tonight so that we can merge.

mzbenami:
looks good, thanks.

youngthejames:
looks good.

PR #48, opened 2015-04-08
Clean up display of test cases

alexthemark:
Cleans up how system tests are displayed. They now look somewhat nicer like JUnit. 

@Aalk4308, I've modified test.bat in a way I think should work in the Windows .bat file, but please test.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -124,5 +124,6 @@
         <delete file="${src.dir}/WordsVal.java" />
         <delete file="${src.dir}/Yylex.java" />
         <delete file="${src.dir}/Yylex.java~" />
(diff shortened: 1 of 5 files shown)


akallem:
This is great, I'm good with this.

alexthemark:
I made it so the shell script will only print the file name instead of the path... Unfortunately I don't know how to do this in batch.

mzbenami:
looks good to me, thanks guys.

akallem:
I'll see if I can have it omit the path in the Windows batch script.

akallem:
OK, I think this is good to go now.

PR #55, opened 2015-04-10
Make log files appear as binary files

alexthemark:
words.log files will no longer appear in diffs.


alexthemark:
Doesn't actually work on github

PR #57, opened 2015-04-10
Local scope

### LINKED ISSUE #50: Implement Local Scope
First version, should be successful for iteration.

alexthemark:
We're getting closer to being a real programming language now. :)
Closes #50 

Implemented as described this morning.

### DIFF
--- a/src/words/environment/WordsEnvironment.java
+++ b/src/words/environment/WordsEnvironment.java
@@ -77,6 +77,24 @@ public class WordsEnvironment {
 	}
 	
 	/**
+	 * Get the number of different object scopes at the moment. Includes the global scope.
+	 */
+	public int getNumberOfScopes() {
+		return gettableObjects.size();
+	}
+	
+	/**
+	 * Pop local scopes off the table until you arrive at the correct number.
+	 * Useful for resetting the symbol table after an error is thrown.
+	 */
+	public void popScopesUntilSize(int desiredSize) {
+		assert gettableObjects.size() >= desiredSize : "Symbol table is smaller than desired size";
+		while (gettableObjects.size() > desiredSize) {
+			exitLocalScope();
+		}
+	}
+	
+	/**
 	 * Creates a new object in the environment and returns it.  Throws an exception if the object could not be created.
 	 * A new object is always added to the most local scope in use
 	 */
--- a/src/words/environment/WordsEnvironment.java
+++ b/src/words/environment/WordsEnvironment.java
@@ -84,17 +84,6 @@ public class WordsEnvironment {
 	}
 	
 	/**
-	 * Pop local scopes off the table until you arrive at the correct number.
-	 * Useful for resetting the symbol table after an error is thrown.
-	 */
-	public void popScopesUntilSize(int desiredSize) {
-		assert gettableObjects.size() >= desiredSize : "Symbol table is smaller than desired size";
-		while (gettableObjects.size() > desiredSize) {
-			exitLocalScope();
-		}
-	}
-	
-	/**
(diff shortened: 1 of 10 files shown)


src/words/environment/WordsEnvironment.java
```diff
@@ -10,13 +12,21 @@
...
 public class WordsEnvironment {
 	private HashMap<String, WordsClass> classes;
-	private HashMap<String, WordsObject> objects;
+	/* gettableObjects is a list of locally scoped symbol tables with the most local 
+	 * scope first in the list, and global scope always being last in the list.
+	 */
+	private LinkedList<HashMap<String, WordsObject>> gettableObjects;
+	/* ungettableObjects contains the objects that have gone out of scope, but should still appear
+	 * on the board, and are still used by action listeners */
+	private HashSet<WordsObject> ungettableObjects;
```
akallem, line 21: Is the advantage of using a HashSet here that, later on when we need to remove an object from all our data structures that reference it, we can do so directly?  As it is now, it seems we don't use the fact that ungettableObjects is a HashSet versus some other structure.

akallem:
Overall this is quite clever and looks very good.

On the tests, could we also have one that tests that an outer scope object is accessible from inner scope (and that this leads to either successful creation or the ObjectAlreadyExists exception)?  I don't see that being tested directly right now.

test/words/test/TestINodeRepeat.java
```diff
@@ -32,4 +33,25 @@ public void testBadRepeat() throws WordsRuntimeException {
...
+	}
+	
+	@Test
+	public void scopeHandlesExceptions() throws WordsRuntimeException {
+		AST statementList = new INodeStatementList(createObjectFred);
+		AST iterativeLoop = new INodeRepeat(fiveLeaf, statementList);
+		loop.enqueueAST(createObjectFred);
+		loop.fastForwardEnvironment(1);
+		// this iterative loop will fail 5 times
+		loop.enqueueAST(iterativeLoop);
```
akallem, line 55: Are you sure this test is working?  It looks to me like you're not fast forwarding the environment to actually run the loop, so the exceptions of trying to create Fred when he already exists in outer scope are never actually generated

akallem:
The one thing I'm not clear on is how scopes are popped when an exception occurs.  See my line comment, I don't think that test is working correctly.

In principle, wouldn't any AST node that caused the environment to enter a new scope need to catch any exceptions that occur within that scope, pop the scope, and then rethrow the exception upwards?

src/words/environment/WordsEnvironment.java
alexthemark, line 21: Yep. The whole advantage is when it comes to removal, we have a constant time removal (provided we have a reference to the object).

test/words/test/TestINodeRepeat.java
alexthemark, line 55: Now it fast forwards.

alexthemark:
I realized that scopes don't need to be explicitly handled at the exception level (at least for the repeat node). 

The scope is entered and exited on either side of a statement list (INodeRepeat.java, lines 23-25), and each exception is handled at the statement level. If an exception is thrown, it will print to the console from  the statement list evaluated on INodeRepeat.java, line 24. An exception should never bubble up to mess up scope.

src/words/environment/WordsEnvironment.java
akallem, line 21: Sweet.

akallem:
Ah, I see what you mean, you're right.  I was thinking any exception would bubble up to the statement list that the repeat statement is a part of, but you're right, it would only bubble up to the statement list contained within the repeat statement, so the scopes are still correct.

akallem:
This looks good to me.

PR #58, opened 2015-04-10
Number Labels on GUI Grid

### LINKED ISSUE #49: Add row and column numbers to GUI
We had discussed adding this last Monday. I'm hoping it can be accomplished quickly and without much effort.

akallem:
Currently, positive y goes down, consistent with graphics and with how one things of rows in a table (row 2 is below row 1).

Closes #49.

### DIFF
--- a/src/words/WordsUI.java
+++ b/src/words/WordsUI.java
@@ -26,8 +26,11 @@ public class WordsUI {
 	static final int moveStep = 2;
 	int xCenterCell, yCenterCell;		// The board coordinates of the cell depicted in the center
 
-	static final int topPadding = 20;	// Padding at the top of the window before the board in pixels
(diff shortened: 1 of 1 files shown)


alexthemark:
I know that it's consistent with computer graphics to have positive y go down. But it's inconsistent with almost all coordinate systems that our users are likely to be familiar with. I feel pretty strongly that we should have positive y going up.

akallem:
This might be one of those 50/50 decisions we shouldn't stress about right now.  As it is, the labels in the GUI just follow the coordinate system that's already in place in the code (e.g., MoveAction down causes y to go up).  So let's revisit the coordinate system displayed to the user in a separate discussion/branch.

mzbenami:
for the future discussion: agree with alex, kids are likely to be used to the axis layouts they see in math class.

PR #59, opened 2015-04-10
WordsClass Tests

akallem:
This branch includes an implementation of setProperty in the WordsClass class, and also implements units tests for WordsClass (including tests of getting/setting properties and class inheritance).


alexthemark:
Looks good to me.

mzbenami:
nice!

PR #63, opened 2015-04-11
add AST to WordsSay

### LINKED ISSUE #16: Say doesn't take an AST.
Make it so Say can take an AST.

zwdnyc:
resolve #16 
1. add AST to WordsSay
2. move WordsProgramExceptions to words.exceptions

### DIFF
--- a/src/words/environment/WordsSay.java
+++ b/src/words/environment/WordsSay.java
@@ -9,15 +9,10 @@ import words.exceptions.WordsProgramException;
 import words.exceptions.WordsRuntimeException;
 
 public class WordsSay extends WordsAction {
-	private AST messageExpression;
-	private String messageValue;
+	private AST message;
 
-	public WordsSay(AST messageExpression) {
-		this.messageExpression = messageExpression;
-	}
-
-	public WordsSay(String messageValue) {
-		this.messageValue = messageValue;
+	public WordsSay(AST message) {
+		this.message = message;
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


alexthemark:
Do you know what's causing your blank lines to appear weird in the diff? It makes it hard to read the PR. Otherwise, looks good to me.

zwdnyc:
My editor auto deletes trailing whitespaces while saving. Sorry for any inconvenience.

Simply adding ?w=1 to the end of the URL will cause GitHub to ignore whitespace for all the diffs on the page.

src/words/environment/WordsSay.java
```diff
@@ -1,22 +1,49 @@
...
-	public WordsSay(String message) {
-		this.message = message;
+	private AST messageExpression;
+	private String messageValue;
+
+	public WordsSay(AST messageExpression) {
+		this.messageExpression = messageExpression;
+	}
+
+	public WordsSay(String messageValue) {
```
akallem: I don't think we should allow this form of construction anymore.  This was the old, incorrect way of doing it.  The grammar only allows an expression, so I think we should do the same here.  That means some existing tests will need to be modified, but that's OK and should be easy to change.

src/words/environment/WordsSay.java
```diff
@@ -1,22 +1,49 @@
...
 	@Override
 	public boolean isExecutable() {
 		return true;
 	}
 
 	@Override
-	protected void doExecute(WordsObject object, WordsEnvironment environment) {
-		object.setMessage(message);
+	protected void doExecute(WordsObject object, WordsEnvironment environment) throws WordsProgramException {
+		if (messageExpression != null) {
```
akallem: Per above, I think this should be changed that the only way to construct it is with an expression, in which case this check isn't necessary (but we could have an assertion for it).

alexthemark:
Cool trick on the whitespace, thanks!

zwdnyc:
WordsSay only takes AST now.

akallem:
OK, this looks good to me now.

mzbenami:
looks good to me, too.

PR #64, opened 2015-04-11
Test INodeQueueSay

### LINKED ISSUE #37: Test evalQueueSay
Using JUnit

zwdnyc:
resolve #37


akallem:
This looks good.  The one thing I'd suggest adding is a test that the right exception is thrown when we try to have an object say something that is not a string or number.  But I suppose that test could either be here or in the WordsSay class.

mzbenami:
lgtm

zwdnyc:
Added a test for expected WordsInvalidTypeException when an object says something other than a string or a number.

akallem:
This looks good to me now.

Issue #52, opened 2015-04-10
Implement Exit, Reset


alexthemark:
Implement the nodes that implement Exit and Reset.

PR #65, opened 2015-04-12
Reset exit

mzbenami:
this is pretty simple but i think it gets the job done. the environment's resetEnvironment() method may have to be updated in the future, especially as we get into event listeners. as of now, resetEnvironment() clears all the collections, but I don't know of a way to truly delete an object from memory or remove all references to an object. testing for the INodeReset class will also have to be extended in the future as we get into classes and event listeners. system tests for both included in the commit.

is it possible to write a JUnit test for INodeExit?


system_test/ResetTest.words.log
```diff
@@ -0,0 +1,500 @@
+> > frame #: 1
```
akallem: This log doesn't look right: it should be showing where the object(s) are and what they're saying.

EDIT: Ah, the reason is that the Reset happens immediately in the first frame, so there is no frame in which Alex is ever in the position.  I guess right now there's nothing we can do about that, since we can't delay the Reset (e.g., using an event listener or custom action).  Perhaps, then, the better test would be to create an object, reset, then create a differently-named object with a different message in a different position.  Then at least the log will show where that new object is, instead of being empty.

src/words/ast/INodeExit.java
```diff
@@ -10,7 +10,8 @@ public INodeExit(Object... children) {
 
 	@Override
 	public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException {
-		// TODO
-		throw new AssertionError("Not yet implemented");
+		System.exit(0);
```
akallem, line 13: Awesome!

system_test/ResetTest.words.log
akallem: Also, the first line of this log looks odd with the two > characters.  That's different from the other logs.  I'm not sure if something went wrong when it was generated?

system_test/ResetTest.words
```diff
@@ -0,0 +1,12 @@
+Alex is a thing at 0,0.
+Alex's name is "alex".
+Make Alex say Alex's name.
+
+Reset.
+
+Alex is a thing at 0,0.
+Make Alex say Alex's name.
```
akallem: This line will generate a runtime error, since right now an object can't say something that is NOTHING, and Alex's name is nothing at this point.  When I run the system_tests, the syntax error is printed, which is confusing.  Right now we should probably have all our system_test programs run without errors, and we'll figure out separately how to write tests that runtime errors are generated when we expect them to be.

akallem:
It looks like there's a JUnit extension here that would allow us to test that Exit works.  http://stefanbirkner.github.io/system-rules/#ExpectedSystemExit.  But for that one, it's so simple I think maybe we can all agree that it'll be OK not to test it via a unit test.  We could probably do it as a system test -- the log will just be shorter than 500 lines (to wherever it exited), and that will suffice.

mzbenami:
I changed the Reset system test (and accompanying log) to what Andrew suggested.

mzbenami:
Closes #52

mzbenami:
closes #52

akallem:
Looks good to me.

PR #82, opened 2015-04-14
System Test Refactor

akallem:
This branch renames the system test programs and log files to use a consistent naming, and it reduces the log files length from 500 to just 100.

(It also incorporates the flip y-axis branch.)


akallem:
@alexthemark @mzbenami @ZhangWangda @jtyoung1292 to make sure any future system tests we write are consistent with this, it would be great it you could review this soon so we can merge it soon (if you agree it makes sense).

alexthemark:
As I said in the meeting, I'd prefer to do this but also have an option

alexthemark:
Oops, accidentally closed and reopened... never mind that.

I'm good with this change. I thought more about my idea of having max frames be an argument into Words, and that would be really annoying since we've automated all of our tests and haven't provided a good location where we could include that argument. 100 frames seems reasonable.

zwdnyc:
looks nice

akallem:
Thanks, glad you thought about it more, because that was exactly going to be my response.  :-)

PR #83, opened 2015-04-14
Enable java assertions by default for tests

zwdnyc:
Hello guys,

I found that assertions in java are not enabled by default. So I suggest that we enable assertions in unit tests and system tests. Previously, the java assertions (not the junit assertions) do not really work.

I was reading the WordsAction.java, and realized that the assertions for isExecutable() and isExpandable() do not work. In fact, they appear to be incorrect and I just fixed them.

Please take a look and see if we should merge this asap.

Thanks,
Wangda

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -88,6 +88,9 @@
     <target name="test-unit" depends="clean, test-compile">
         <mkdir dir="${test.report.dir}" />
         <junit printsummary="on" fork="true" haltonfailure="true">
+            <assertions>
+                <enable />
+            </assertions>
             <classpath refid="junit.classpath" />
             <formatter type="plain" />
             <batchtest todir="${test.report.dir}">
--- a/src/words/ast/INodeAssign.java
+++ b/src/words/ast/INodeAssign.java
@@ -17,9 +17,8 @@ public class INodeAssign extends INode {
 	@Override
 	public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException {
 		WordsObject obj = children.get(0).eval(environment).objValue;
-		if (obj == null) {
-			throw new AssertionError("Obj should never be null here");
-		}
+		assert obj != null : "Obj was null when it shouldn't have been.";
+		
 		String propertyName = children.get(1).eval(environment).stringValue;
 		ASTValue propertyASTValue = children.get(2).eval(environment);	
 
--- a/src/words/ast/INodeRetrieveProperty.java
+++ b/src/words/ast/INodeRetrieveProperty.java
@@ -27,10 +27,8 @@ public class INodeRetrieveProperty extends INode {
 		}
 		
 		WordsObject obj = refList.objValue;
-		if (obj == null) {
-			throw new AssertionError("obj shouldn't be null");
-		}
-		
+		assert obj != null : "Obj was null when it shouldn't have been.";
+
 		ASTValue id = children.get(1).eval(environment);
 		String propName = id.stringValue;
 		
--- a/src/words/environment/WordsAction.java
+++ b/src/words/environment/WordsAction.java
@@ -15,8 +15,7 @@ public abstract class WordsAction {
 	 * Execute the action if it is expandable.
 	 */
 	public final void execute(WordsObject object, WordsEnvironment environment) throws WordsProgramException {
-		boolean flag = !isExecutable();
-		assert flag : "Attempted to execute non-executable action";
+		assert isExecutable() : "Attempted to execute non-executable action";
 		doExecute(object, environment);
 	};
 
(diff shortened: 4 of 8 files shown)


mzbenami:
i'm fine with this change but would someone mind explaining why sometimes we use this assert syntax and sometimes we use throw new AssertionError()? Are both not enabled by default?

zwdnyc:
I think "throw new AssertionError()" always works. To use "assert", we need the -enableassertions switch.

In my opinion, we'd better (almost) always use this assert syntax for testing the code and do not explicitly throw AssertionErrors, since these assertions can then be turned off easily and we'll have no overhead running product code.

See http://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html for an example where using throw new AssertionError() may be favorable in the default case of a switch statement.

akallem:
Wow, great catch Wangda!  Hadn't even noticed that they weren't enabled by default.  And thanks also for fixing those isExecutable/isExpandable assertions; yes, I'd written them backwards (know idea why!).

I agree with you about using assert in favor of throwing AssertionError, and as you said, that the only place we should explicitly throw an AssertionError is the default case of a switch, which I believe is what we're doing now.

I'm good with this change and agree we should merge ASAP.

alexthemark:
Let's get this merged.

alexthemark:
We should also probably write tests that actually use our assertions. If we had tested a single assertion, we would have noticed this.

akallem:
I disagree.  I think our asserts are just a defensive coding practice to
help us debug errors more easily.  Had we never heard of asserts and never
used them, we wouldn't be discussing testing them.

In other words, the assert is not a required functional behavior of our
code that we need to validate works by testing it.  Instead, it's just a
defensive coding practice to help us debug errors if they arise.

alexthemark:
Yeah, but we've written asserts and they're a part of our code base, and we're expecting them to fail. It's defensive, but it's also a defense that we're expecting to work. 

I think it's best to err on the side of, "If it's important enough for us to write into our code base, it's important enough to test that it works."

akallem:
I disagree again.  An assert is essentially just documentation about the logic of our code, but it's documentation in the form of an assert as a way to catch a mistake if it turns out we're wrong.

Many asserts are not even testable.  Consider line words.ast.INodeEquals:61.  There is no way to reach this line without the lhs and rhs to have the same type (assuming the code above it is correct).  The reason the assert is there is defensive: if someone happens to change the code above that causes the assert to fail, we'll know it.  I don't see how you could call INodeEquals.eval() right now to cause that assert to be triggered.

mzbenami:
i don't like it when mommy and daddy fight. but i agree with andrew.
also, before we merge this, all throw new AssertionError()s should be changed over to this syntax.

akallem:
Hahaha, as long as I'm the daddy and not the mommy, I'm OK with that analogy. ;-)

By the way, another reason why it's not right to unit test our assert statements is that unit tests should (I think) test that a method meets its specification, i.e. that the method gives the expected output for all possible classes of input.  A method specification never says the method should throw assertion exceptions.  If the writer wants to use assertion errors to help implement and future-proof their method, that makes sense.  And if an assertion can be triggered, some existing unit tests (which in theory cover all possible types of input) should trigger it.  There's thus no need to test the assertion separately (or way to test it, if the method is implemented correctly).

@mzbenami Do we have many/any cases where we are directly throwing an AssertionError, besides as the default case in a switch (where it's appropriate to do so)?

mzbenami:
yes:
- line 21 of INodeAssign.java
- line 31 of INodeRetrieveProperty.java

both of those are things that i did. i can fix them on my next branch, no big deal.

akallem:
OK.  But rather than fix them on your next branch, I'd suggest pulling this branch, fixing them here, and then pushing back to this branch, since it's related.

mzbenami:
done.

akallem:
OK, let's get this merged.  @alexthemark @jtyoung1292 what are your thoughts?

alexthemark:
I'll meet you halfway and say that certain asserts should be tested. 

For instance, WordsEnvironment.exitLocalScope() has an assertion to make sure that the global stack has not been popped. Maybe this is a bad case for an assert and it should be another type of error (i.e. WordsScopeException or something new). In fact, I'm reading over assert documentation right now and it seems like I incorrectly used assert there.

alexthemark:
But yeah, this branch can be merged.

akallem:
OK cool.  Wangda, merge away.  :-)

Separately, how are you using assert incorrectly there?  It looks OK to me.  And I think it is appropriate as an assert and not a user-level exception.  The user should have no possible way to trigger the global scope being popped.  If they do, we've done something wrong.

alexthemark:
It's not words user level -- if we ever pop the global scope, it's because one of us screwed up, and we probably want the program to crash (which the assert causes). However, it is argument checking a public method of a class, which is explicitly frowned upon here http://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html

And then, since it is a behavior of a public method, it should be tested (not that this really matters too much -- I'm going to spend my time on the basic action predicate and not on changing this exception).

PR #86, opened 2015-04-15
Delay Queue Move Eval

### LINKED ISSUE #85: Queue Move Not Delaying Evaluation of Distance
The reference manual says the distance should be evaluated only when the move action is expanded, but the current implementation is evaluating the distance immediately at enqueue time.

akallem:
This branch delays the evaluation of the distance of a move until it is expanded.

For the moment, I've included a simple system_test, which will need to be further fleshed out.  So this branch is not quite ready to be merged just yet.

@ZhangWangda , I'm sharing this now primarily because I think it may be helpful when implementing WordsWait to do the same delayed evaluation.  (Similar to what you already did for WordsSay.)

Closes #85

### DIFF
--- a/src/words/ast/INodeQueueMove.java
+++ b/src/words/ast/INodeQueueMove.java
@@ -1,6 +1,5 @@
 package words.ast;
 
-import words.ast.ASTValue.ValueType;
 import words.environment.*;
 import words.exceptions.*;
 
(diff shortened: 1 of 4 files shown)


zwdnyc:
ok, thanks

akallem:
I'm closing this as a pull request, because I'm including this in another branch I'm working on.  But I'll leave the branch up in case you want to take a further look.

PR #87, opened 2015-04-16
Queued Property Assignment

### LINKED ISSUE #70: Queued Property Assignment
Queue property assignments

### LINKED ISSUE #85: Queue Move Not Delaying Evaluation of Distance

akallem:
This branch has an implementation of queued property assignment (statements like "Make Fred's height be 68, weight be 150, age be 30.").  It also includes an implementation for inherited attributes in our AST (discussed more below).  Up to know, we've only needed synthesized attributes (INodeAdd returns a sum, etc.), but here is a case where I believe the proper implementation is with an inherited attribute.  I think inherited attributes will also be useful for class definition statements.

Closes #70 
Closes #85 

To understand the implementation of queued property assignment, the grammar defines three INodes:
- INodeQueueAssign: Corresponds to the entire example statement above
- INodeQueueAssignPropertyList: Corresponds to "height be 68, weight be 150, age be 30" above
- INodeQueueAssignProperty: Corresponds to "height be 68" or the other two above

Now, consider INodeQueueAssignProperty's implementation of eval().  It needs to know what object to set the property on.  But obviously, it doesn't know that from its children (its children are just the property name and property value expression).  The object to set the property on is known only to the grandparent node.  Thus, INodeQueueAssignProperty must inherit that WordsObject from the grandparent.

To implement that, I added a second eval() method to the abstract INode class which takes an Object (in the Java sense) parameter called inherited.  The default is just to ignore any passed inherited attribute and call the regular eval() implementation.  However, INode subclasses can override the default if they want to use the inherited attribute.  In their implementation, they cast the Java Object into whatever they know it was that their caller passed in (which could be a HashMap if they needed to inherit multiple properties, though I doubt we'll need that).

You can see this in action in the doExecute() method of WordsPropertyAssignment.  Here, I call the propertyAssignmentList's eval() method, passing the WordsObject as the inherited attribute.

Separately, I noticed that INodeQueueMove and WordsMove was evaluating the distance expression immediately instead of when the WordsMove action is expanded.  I fixed that on a separate branch merged into this one and showing as a single pull request.

Two other minor things to note about this branch:
- There is now more than one place where we need to take an ASTValue of NUM, STRING, OBJ, or NOTHING type and convert it to a corresponding WordsProperty.  So I factored that out and made it a method of ASTValue.
- I added some additional tests for INodeWordsMove

### DIFF
--- a/src/words/ast/ASTValue.java
+++ b/src/words/ast/ASTValue.java
@@ -3,6 +3,7 @@ package words.ast;
 import words.environment.Direction;
 import words.environment.WordsObject;
 import words.environment.WordsPosition;
+import words.environment.WordsProperty;
 
 /**
  * A dynamically-typed value that an AST node can return as its result.
@@ -90,5 +91,25 @@ public class ASTValue {
 		}
 		
 		return this;
+	}
+	
+	/**
+	 * Returns a WordsProperty object of type NUM, STRING, OBJ, or NOTHING
+	 * corresponding to this ASTValue.  Calling on an ASTValue with a type
+	 * other than those listed above is prohibited.
+	 */
+	public WordsProperty toWordsProperty() {
+		switch (this.type) {
+			case NUM:
+				return new WordsProperty(this.numValue);
+			case STRING:
+				return new WordsProperty(this.stringValue);
+			case OBJ:
+				return new WordsProperty(this.objValue);
+			case NOTHING:
+				return new WordsProperty(WordsProperty.PropertyType.NOTHING);
+			default:
+				throw new AssertionError("Cannot convert ASTValue of type " + this.type.toString() + "to WordsProperty");
+		}
 	}
 }
--- a/src/words/ast/AST.java
+++ b/src/words/ast/AST.java
@@ -16,11 +16,22 @@ public abstract class AST {
 	 * 
 	 * @param level the indentation level
 	 */
-	abstract public void dump(int level);
+	public abstract void dump(int level);
 	
 	/**
 	 * Evaluate an AST node to return an ASTValue and possibly have side effects on the passed environment.
+	 * 
 	 * @throws WordsRuntimeException 
 	 */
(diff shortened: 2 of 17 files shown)


src/words/ast/ASTValue.java
```diff
@@ -91,4 +92,24 @@ public ASTValue tryCoerceTo(ValueType newType) {
 		
 		return this;
 	}
+	
+	/**
+	 * Returns a WordsProperty object of type NUM, STRING, OBJ, or NOTHING
+	 * corresponding to this ASTValue.  Calling on an ASTValue with a type
+	 * other than those listed above is prohibited.
+	 */
+	public WordsProperty toWordsProperty() {
```
alexthemark, line 101: Solid refactor.

alexthemark:
Looks good to me.

Having the ability to send an object down the tree is going to be very useful in a few other sections -- I have a feeling that this is going to be very helpful when we're doing aliasing for basic action predicates.

akallem:
Here's a quick question for the group.  Right now, a queued property assignment action eats up a frame when it executes.  The object will just sit there as if it were executing a wait action, but behind the scenes the property is updated.  Would it be better (clearer to users) if property assignment didn't eat up a frame?  That way, in a snippet like below, the object would never step moving.

Make Fred move up 5.
Make Fred's status be "halfway".
Make Fred move down 5.
Make Fred's status be "done".

mzbenami:
looks good to me, as well.

alexthemark:
I think that we should continue to have property assignment take up a frame. This keeps it consistent that anything that is queued takes one frame. I don't like special cases in the queue.

mzbenami:
as to your question, i agree with alex. we live by the queue, we die by the queue.

akallem:
OK, that's fine with me.  :-)

PR #88, opened 2015-04-16
Class create

### LINKED ISSUE #67: Class Basic Creation
Should handle assignment of properties but not custom actions.

mzbenami:
Hey y'all

This commit touches the Environment class, INodeCreateClass, and INodeDefineProperty so that classes can be created. System tests look good.

What do you guys think about removing

class_definition_statement:
        HAS A identifier '.'

from the grammar? In reality it's no-op. Nothing changes in the environment from this line from what I can tell. This commit supports it, though, and does what is expected.

Another thing, I still have JUnit tests to write, but want to check to see if I should run AST tests, Environment tests, or both. I don't see AST tests for INodeCreateObject, and for class creation, the AST tests and environment tests would be very similar. Both would check results by accessing the environment directly. The only difference is in the entry point for class creation. Let me know what you think.

Closes #67

### DIFF
--- a/src/words/ast/INodeCreateClass.java
+++ b/src/words/ast/INodeCreateClass.java
@@ -24,8 +24,6 @@ public class INodeCreateClass extends INode {
 		if (statementList != null) {
 			for (Object child : statementList.children) {
 				if (child instanceof INodeDefineProperty) {
-					
-					assert (((INodeDefineProperty) child).id == null) : "ID WAS NOT NULL";
 					ASTValue propertyName = ((INodeDefineProperty) child).id.eval(environment);
 					
 					if (((INodeDefineProperty) child).literal == null) {
--- a/src/words/ast/INodeCreateClass.java
+++ b/src/words/ast/INodeCreateClass.java
@@ -1,11 +1,7 @@
 package words.ast;
 
-import words.ast.ASTValue.ValueType;
 import words.environment.WordsClass;
 import words.environment.WordsEnvironment;
-import words.environment.WordsProperty;
-import words.exceptions.WordsClassAlreadyExistsException;
-import words.exceptions.WordsClassNotFoundException;
 import words.exceptions.WordsRuntimeException;
 
 public class INodeCreateClass extends INode {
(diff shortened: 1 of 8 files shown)


akallem:
I agree it makes sense to remove "HAS A identifier" from the grammar.  I think it was originally there because, at the time, we were thinking some properties could be booleans.  But we ended up not pursuing that.

src/words/ast/INodeCreateClass.java
```diff
@@ -10,7 +15,40 @@ public INodeCreateClass(Object... children) {
...
+		INodeClassStatementList statementList = (INodeClassStatementList) this.children.get(2);
+		if (statementList != null) {
+			for (Object child : statementList.children) {
+				if (child instanceof INodeDefineProperty) {
+					ASTValue propertyName = ((INodeDefineProperty) child).id.eval(environment);
+					
+					if (((INodeDefineProperty) child).literal == null) {
+						wordsClass.setProperty(propertyName.stringValue, new WordsProperty(WordsProperty.PropertyType.NOTHING));
+					} else {
+						ASTValue propertyValue = ((INodeDefineProperty) child).literal.eval(environment);
```
akallem: I've implemented a method that simplifies this.  It happens to be on the queue-prop-assign-branch in #87.  I think it would be good if you merged that branch into this one and used that refactor.  (Sorry that it's not on a separate branch by itself, which would have been better practice on my part.)

src/words/ast/INodeCreateClass.java
```diff
@@ -10,7 +15,40 @@ public INodeCreateClass(Object... children) {
...
-		// TODO
-		throw new AssertionError("Not yet implemented");
+		ASTValue className  = this.children.get(0).eval(environment);
+		ASTValue parentClassName = this.children.get(1).eval(environment);
+		
+		WordsClass wordsClass = environment.createClass(className.stringValue, parentClassName.stringValue);
+		
+		INodeClassStatementList statementList = (INodeClassStatementList) this.children.get(2);
+		if (statementList != null) {
+			for (Object child : statementList.children) {
```
akallem: This part would be an opportunity to leverage the inherited attributes component of the queue-prop-assign branch in #87, so that the implementation of INodeClassStatementList is separated out (and it would receive the class object is an inherited attribute).

akallem:
Overall this looks great!  I added two line comments where I think you could incorporate some stuff from my latest branch.

I don't have a strong opinion on what's the right class to unit test here.  I agree that it would be very similar.

mzbenami:
OK, I will do the JUnit testing on the environment to keep it consistent. I'll also try to merge in the refactors you suggested.

mzbenami:
Ok guys, I think this is good to go. After Andrew's refactoring the JUnit tests for the INodes really looked trivial so I just put a bunch of tests for the environment in the "Class Creation" section.

PR #89, opened 2015-04-16
Implement wait

### LINKED ISSUE #73: Implement Wait

zwdnyc:
resolve #73 

this branch implements wait, with unit tests for INodeQueueWait and WordsWait

this branch also fixes some minor problems with Say


src/words/ast/INodeQueueSay.java
```diff
@@ -31,6 +32,9 @@ public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException
 		} else {
 			object = environment.getObject(identifier.stringValue);
 		}
+		if (object == null) {
+			throw new WordsObjectNotFoundException(identifier.stringValue);
```
alexthemark: WordsEnvironment.getObject() should throw this error if the object isn't found.

src/words/ast/INodeQueueWait.java
```diff
@@ -10,7 +17,33 @@ public INodeQueueWait(Object... children) {
...
+		if (referenceObject.type.equals(ASTValue.ValueType.OBJ)){
+			WordsProperty property = referenceObject.objValue.getProperty(identifier.stringValue);
+			if (property.type != WordsProperty.PropertyType.OBJECT) {
+				throw new WordsInvalidTypeException(ASTValue.ValueType.OBJ.toString(), property.type.toString());
+			}
+			object = property.objProperty;
+		} else {
+			object = environment.getObject(identifier.stringValue);
+		}
+		if (object == null) {
```
alexthemark: Same comment as above... we should never have a null object coming out of environment.getObject. This exception is thrown by the environment.

src/words/ast/INodeQueueSay.java
zwdnyc: OK, but can object = property.objProperty; result in null object? I see in INodeQueueMove we also have this check.

akallem: If property is of type WordsProperty.PropertyType.OBJECT, then property.objProperty cannot be null.

I agree this check can be removed (and in INodeQueueMove).  I think it's legacy code from when environment.getObject() returned null when an object wasn't found instead of throwing an exception.  We changed it to throw an exception but didn't change it here.

src/words/ast/INodeQueueWait.java
akallem: This pattern of code from lines 25:34 appears in several places.  I'm going to open an issue to factor this out as a separate method.  (That doesn't need to be done on this branch, we can address it in our next sprint.)

akallem:
Aside from the object=null check that can be removed, everything else here looks great to me.

zwdnyc:
those checks are removed now.

mzbenami:
cool!

akallem:
Just thought of one thing: we should probably have at least a simple system test that incorporates wait.  Maybe just add it as a line in BasicTest.words, since it's pretty core functionality?

alexthemark:
I'm a fan of incorporating wait into BasicTest. I'm merging this into my branch for basic-action-predicate so I can test waits.

akallem:
Sweet.

PR #92, opened 2015-04-16
Implement event listeners for boolen predicates

### LINKED ISSUE #69: Listener Creation
Basic creation, and eval each frame for a boolean

zwdnyc:
resolve #69 

Please take a look and see if this implementation is the same as what the language manual describes.


src/words/environment/WordsEventListener.java
```diff
@@ -1,15 +1,35 @@
...
+	public boolean execute(WordsEnvironment environment) throws WordsRuntimeException {
+		ASTValue predicateValue = predicate.eval(environment);
+
+		// currently only restricted to boolean predicate
+		assert predicateValue.type == ASTValue.ValueType.BOOLEAN : "Predicate has type " + predicateValue.type.toString();
+
+		if (predicateValue.booleanValue == true) {
+			statementList.eval(environment);
+		} else {
+			if (temporary)
```
alexthemark: This lack of braces doesn't follow our style guide.

src/words/FrameLoop.java
```diff
@@ -87,8 +88,17 @@ private boolean executeSingleFrame() {
...
-			eventListener.execute();
+		for (Iterator<WordsEventListener> iterator = environment.getEventListeners().iterator(); iterator.hasNext();) {
+			try {
+				WordsEventListener eventListener = iterator.next();
+				boolean delete = !eventListener.execute(environment);
+				if (delete) {
+					iterator.remove();
+				}
+			} catch (WordsRuntimeException e) {
+				System.err.println(e.toString());
```
alexthemark, line 99: If we directly print a runtime exception, we're not giving the user any information about where the program went wrong (only what went wrong). By here, I think this should be a WordsProgramException.

src/words/environment/WordsEventListener.java
```diff
@@ -1,15 +1,35 @@
...
+	public WordsEventListener(AST predicate, AST statementList, boolean temporary) {
+		this.predicate = predicate;
+		this.statementList = statementList;
+		this.temporary = temporary;
+	}
 	
-	public void execute() {
-		// TODO Auto-generated method stub
-		
+	public boolean execute(WordsEnvironment environment) throws WordsRuntimeException {
```
alexthemark: I think that this should probably throw a WordsProgramException, since this is where the AST lives.

zwdnyc: throws WordsProgramException now.

src/words/FrameLoop.java
```diff
@@ -87,8 +88,17 @@ private boolean executeSingleFrame() {
 		}
 
 		// Phase 3: Listener Evaluation
-		for (WordsEventListener eventListener : environment.getEventListeners()) {
-			eventListener.execute();
+		for (Iterator<WordsEventListener> iterator = environment.getEventListeners().iterator(); iterator.hasNext();) {
```
akallem, line 91: Could we use the iterator shorthand like this?  Is there a reason to prefer the more verbose form?

for (WordsEventListener : evenListener : environment.getEventListeners()) {

}

Is it because we may be deleting the listener during the loop?

src/words/FrameLoop.java
```diff
@@ -87,8 +88,17 @@ private boolean executeSingleFrame() {
...
+			try {
+				WordsEventListener eventListener = iterator.next();
+				boolean delete = !eventListener.execute(environment);
+				if (delete) {
+					iterator.remove();
+				}
+			} catch (WordsRuntimeException e) {
+				System.err.println(e.toString());
+				System.out.println("> ");
```
Aalk4308: I don't really like how these System.out.println("> ") prompts are starting to profilerate all over our code.  I'm going to open an issue for someone to fix that.

src/words/FrameLoop.java
```diff
@@ -87,8 +88,17 @@ private boolean executeSingleFrame() {
...
+		for (Iterator<WordsEventListener> iterator = environment.getEventListeners().iterator(); iterator.hasNext();) {
+			try {
+				WordsEventListener eventListener = iterator.next();
+				boolean delete = !eventListener.execute(environment);
+				if (delete) {
+					iterator.remove();
+				}
+			} catch (WordsProgramException e) {
+				System.err.println(e.toString());
+				System.out.println("> ");
```
akallem, line 100: I don't really like how these System.out.println("> ") prompts are starting to profilerate all over our code. I'm going to open an issue for someone to fix that.

(This repeats a comment I made a moment ago on an old diff.)

akallem:
I'm continuing to take a closer look but so far this looks amazing, I love it!

Issue #77, opened 2015-04-13
Aliases During Event Listeners


alexthemark:
Allow objects to be aliased into a local scope when an event listener fires.

PR #96, opened 2015-04-16
Basic action predicate

### LINKED ISSUE #93: Basic Action Predicates
Forgot to assign this to myself the other day.

alexthemark:
Incorporates Wangda's latest branch for Wait (that way I could make the wait predicate work).

There's a good amount of testing that I still need to do once I can merge in class creation, since these listeners should work with inheritance. It's also incomplete in that there's no aliasing -- however, this should give us a solid start to create the full listeners. 

I also reworked how objects were being stored in the environment. Previously, all "ungettable" objects were stored in one HashSet. Now, objects are always stored by their class, and continue to be stored by their name when they are in scope. Once again, once we have full class creation, I will add more testing to these methods. 

I also renamed "getCurrentCell" to "getCurrentPosition" since we don't ever use the term "cell" anymore. However, I'm now seeing that this may create some merge conflicts. 

Closes #93, #77


mzbenami:
feel free to merge in my class creation branch. only minor changes and junit testing left. the code works.

alexthemark:
I will do that later tonight.

akallem:
Cool, I'm taking a look through now.

Out of curiosity, since you've set up your project in Eclipse where the main source and the JUnit tests are linked together, does that make it possible to use Eclipse's refactor functionality to turn getCurrentCell into getCurrentPosition automatically?  Since my Eclipse project doesn't isn't properly setup for the JUnit tests right now, I couldn't refactor the main source without breaking the tests.

The reason I ask is that we might want to rename some other things.  I don't know about everyone else, but for me the proliferation of the word "Words" sometimes makes my head spin!  I think it's a bit of an accident of history, too.  We initially called our Java classes WordsObject and WordsClass, which was reasonable because we needed to distinguish them from Java objects and classes.  But then we followed that pattern and named pretty much everything else WordsXYZ, even if it wasn't really necessary.

I wonder if our code would be clearer (and shorter) if we refactored "Words" out of a lot of those places.  WordsAction could become Action, WordsEnvironment could become Environment, etc.

src/words/ast/INodeMovesPredicate.java
```diff
@@ -10,7 +15,22 @@ public INodeMovesPredicate(Object... children) {
 
 	@Override
 	public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException {
-		// TODO
-		throw new AssertionError("Not yet implemented");
+		ASTValue className = children.get(0).eval(environment);
```
akallem: In the grammar right now, the "subject" of a basic action predicate can be either a class name _or_ a particular object (using a reference list and identifier).  The first (0 index) child is always of type INodeSubject, but it looks like this implementation assumes it will only return a class name as a string.

src/words/environment/WordsClass.java
```diff
@@ -6,12 +7,18 @@
...
+	private ArrayList<WordsClass> children;
 	
 	public WordsClass(String className, WordsClass parent) {
 		properties = new HashMap<String, WordsProperty>();
 		functions = new HashMap<String, WordsCustomActionDefinition>();
+		children = new ArrayList<WordsClass>();
 		this.parent = parent;
 		this.className = className;
+		if (parent != null) {
+			// only "thing" should get here
```
akallem: I'm confused by this comment.  If parent != null, then shouldn't everything _except_ "thing" get here?

src/words/environment/WordsEnvironment.java
```diff
@@ -15,17 +16,18 @@
 	/* gettableObjects is a list of locally scoped symbol tables with the most local 
 	 * scope first in the list, and global scope always being last in the list.
 	 */
-	private LinkedList<HashMap<String, WordsObject>> gettableObjects;
-	/* ungettableObjects contains the objects that have gone out of scope, but should still appear
-	 * on the board, and are still used by action listeners */
-	private HashSet<WordsObject> ungettableObjects;
+	private LinkedList<HashMap<String, WordsObject>> objectsByName;
```
akallem, line 16: I like this refactor a lot.  I think objectsByName is much clearer than gettableObjects.

src/words/environment/WordsEnvironment.java
```diff
@@ -43,8 +46,7 @@ private void setupEnvironment() {
 	 */
 	public void resetEnvironment() {
 		classes.clear();
-		gettableObjects.clear();
-		ungettableObjects.clear();
+		objectsByName.clear();
```
akallem, line 47: Should there by a line here to also clear objectsByClass?

src/words/environment/WordsEnvironment.java
```diff
@@ -138,12 +140,26 @@ public WordsObject getObject(String objectName) throws WordsRuntimeException {
...
+		try {
+			return getObjectsByClass("thing");
+		} catch (WordsClassNotFoundException e) {
+			System.exit(1);
+		}
+		return null;
+	}
+	
+	/**
+	 * Return a collection of all objects of a certain class
```
akallem: Just for clarity, it might be better to say "of a certain class or any of its subclasses".

src/words/environment/WordsEnvironment.java
```diff
@@ -138,12 +140,26 @@ public WordsObject getObject(String objectName) throws WordsRuntimeException {
 	 * Returns a collection of all objects.
 	 */
 	public Collection<WordsObject> getObjects() {
-		Collection<WordsObject> allObjects = new HashSet<WordsObject>();
-		for (HashMap<String,WordsObject> scopeObjects : gettableObjects) {
-			allObjects.addAll(scopeObjects.values());
+		try {
+			return getObjectsByClass("thing");
```
akallem: Minor side comment: it would probably be good for us to define "thing" as a final constant somewhere called BASE_CLASS rather than using the string directly every time.

src/words/environment/WordsClass.java
alexthemark: Yep, my comment is wrong. will fix.

src/words/ast/INodeMovesPredicate.java
alexthemark: Ah, I didn't realize that we still allowed listeners to be attached to individual objects.  

Ooh, this changes a lot. I will take care of this later tonight.

src/words/environment/WordsEnvironment.java
alexthemark, line 47: Yup. I will clear it.

src/words/environment/WordsEnvironment.java
alexthemark: Will clarify.

src/words/environment/WordsEnvironment.java
alexthemark: I agree. I'll refactor that in.

src/words/ast/INodeMovesPredicate.java
```diff
@@ -10,7 +15,22 @@ public INodeMovesPredicate(Object... children) {
...
+		ASTValue objectAlias = children.get(1).eval(environment);
+		ASTValue moveDirection = children.get(2) == null ? null : children.get(2).eval(environment);
+		
+		HashSet<WordsObject> objectsOfClass = environment.getObjectsByClass(className.stringValue);
+
+		for (WordsObject object : objectsOfClass) {
```
akallem: I have a concern here.  It looks like this implementation causes a Moves predicate to return true whenever _any_ object of the class has moved.  Maybe I haven't thought about it enough, but it seems to me that a basic action predicate needs "return true" _for all_ objects of the class that have moved.  In my mind the predicate needs to be evaluated many times, one for each possible object (or pair of objects, in the case of touches).  In other words, it seems to me that the loop over the objects has to occur _outside_ eval(), not inside, so that we get a different boolean result for each object.

alexthemark:
I agree on refactoring a good number of names, and my Eclipse is set up that it will do so across test code as well. Especially now that we're in a "words" package, we really don't need to prefix each class with it. I'll open an issue for this.

akallem:
Awesome.  Also, I enjoy it when one or more of us is online / on GitHub at the same time.  :-)

src/words/ast/INodeMovesPredicate.java
alexthemark: I had a similar concern, and I was going to bring this up tomorrow. I didn't want to add this functionality before we had listeners, but I think listeners might have to pass down their statement list in the eval function (using the new eval function from #87). Then, inside of this loop, we would create a local scope, run the statement list, and pop the local scope. 

This would result in the statement list being run each time that this predicate is true. I think this is easier than passing the result of the predicate (with scoping issues) up the tree. But I think this might be a good discussion item tomorrow.

akallem: Yeah, I think that's a good way to do it.  For boolean predicates, though, I think the listener itself might have to run its statement list, since boolean predicates appear elsewhere besides listeners.  But for basic action predicates, what you're saying makes sense.  Conceptually, a basic action predicate isn't exactly a predicate but a "template" for a predicate, so it makes sense that a listener would provide the template the code that should be executed for any instantiation of the template that is true.

src/words/ast/INodeBasicActionPredicate.java
```diff
@@ -0,0 +1,13 @@
+package words.ast;
+
+import words.environment.WordsEnvironment;
+import words.exceptions.WordsRuntimeException;
+
+public abstract class INodeBasicActionPredicate extends INode {
```
akallem, line 8: Cool, I like that you made this an abstract class.

src/words/ast/INodeListenerTemp.java
```diff
@@ -13,16 +13,25 @@ public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException
...
+			ASTValue predicateValue = predicate.eval(environment, statementList);
+			
+			if (predicateValue.booleanValue == true) {
+				environment.createListener(predicate, statementList, true);
+			}
+			return null;
+		} else {
+			ASTValue predicateValue = predicate.eval(environment);
+	
+			// currently only restricted to boolean predicate
```
akallem: This comment can be removed.  :-)

src/words/ast/INodeListenerTemp.java
```diff
@@ -13,16 +13,25 @@ public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException
...
-		assert predicateValue.type == ASTValue.ValueType.BOOLEAN : "Predicate has type " + predicateValue.type.toString();
-
-		// temporary listener is created only if the predicate is true
-		if (predicateValue.booleanValue == true) {
-			environment.createListener(predicate, statementList, true);
+		if (predicate instanceof INodeBasicActionPredicate) {
+			ASTValue predicateValue = predicate.eval(environment, statementList);
+			
+			if (predicateValue.booleanValue == true) {
+				environment.createListener(predicate, statementList, true);
```
akallem: I think there's a subtlety here that perhaps is unclear in the language specification.  Do we actually want to execute the statement list right now during this check?  Or just check that the predicate is true, but not evaluate the statement list now (wait until Phase 3 of the frame instead).  I'm concerned we might effectively execute the statement list twice in the same frame.

This could be resolved just by passing null for the statement list, or having a separated eval() method without the inherited attribute that merely checks the predicate but doesn't execute the statements.

src/words/ast/INodeMovesPredicate.java
```diff
@@ -1,16 +1,57 @@
...
+			objectsToCheck = environment.getObjectsByClass(subject.stringValue);
+		} else if (subject.type.equals(ASTValue.ValueType.OBJ)) {
+			objectsToCheck.add(subject.objValue);
+		}
+
+		for (WordsObject object : objectsToCheck) {
+			WordsAction lastAction = object.getLastAction();
+			if (lastAction instanceof WordsMove) {
+				WordsMove lastMove = (WordsMove) lastAction;
+				if (moveDirection == null || moveDirection.directionValue.equals(lastMove.getDirection())) {
```
akallem: The move basic action predicate also allows the 'anywhere' keyword, which in this context is semantically equivalent to the direction being omitted.  So I think this check should also pass if moveDirection is ANYWHERE.  (As it is now, it'll be false because lastMove.getDirection() will only return an explicit direction, never ANYWHERE.)

src/words/ast/INodeMovesPredicate.java
```diff
@@ -1,16 +1,57 @@
...
+		ASTValue objectAlias = children.get(1).eval(environment);
+		ASTValue moveDirection = children.get(2) == null ? null : children.get(2).eval(environment);
+		
+		INodeStatementList stmtList = (INodeStatementList) inheritedStmts;
+		
+		HashSet<WordsObject> objectsToCheck = new HashSet<WordsObject>();
+		ASTValue returnVal = new ASTValue(false);
+		
+		// If subject is a string, then we are looking at a class name
+		if (subject.type.equals(ASTValue.ValueType.STRING)) {
```
akallem: Since you've made INodeBasicActionPredicate an abstract class, this conditional from 34:38 could be separated out as a method that all of the concrete classes use.

src/words/ast/INodeMovesPredicate.java
```diff
@@ -1,16 +1,57 @@
...
+			objectsToCheck.add(subject.objValue);
+		}
+
+		for (WordsObject object : objectsToCheck) {
+			WordsAction lastAction = object.getLastAction();
+			if (lastAction instanceof WordsMove) {
+				WordsMove lastMove = (WordsMove) lastAction;
+				if (moveDirection == null || moveDirection.directionValue.equals(lastMove.getDirection())) {
+					returnVal.booleanValue = true;
+					environment.enterNewLocalScope();
```
akallem, line 37: Lines 46:51 could be separated out into a method in INodeBasicActionPredicate, since all of the concrete subclasses use it.

src/words/ast/INodeTouchesPredicate.java
```diff
@@ -1,16 +1,70 @@
...
+		ASTValue objectAlias1 = children.get(1).eval(environment);
+		ASTValue subject2 = children.get(2).eval(environment);
+		ASTValue objectAlias2 = children.get(3).eval(environment);
+		
+		INodeStatementList stmtList = (INodeStatementList) inheritedStmts;
+		
+		if (objectAlias1.type.equals(ASTValue.ValueType.STRING) 
+				&& objectAlias2.type.equals(ASTValue.ValueType.STRING)
+				&& objectAlias1.stringValue.equals(objectAlias2.stringValue)) {
+			throw new WordsAliasException("Aliases may not be named the same.");
```
akallem: Nice, this is a great check.

src/words/environment/WordsEventListener.java
```diff
@@ -18,29 +19,39 @@ public WordsEventListener(AST predicate, AST statementList, boolean temporary) {
...
-			if (temporary) {
-				return false;
+			ASTValue predicateValue;
+			try {
+				predicateValue = predicate.eval(environment);
+			} catch (WordsRuntimeException e) {
+				throw new WordsProgramException(predicate, e);
 			}
+	
+			// currently restricted to boolean predicate
```
akallem: This comment can be removed.

src/words/environment/WordsEventListener.java
```diff
@@ -18,29 +19,39 @@ public WordsEventListener(AST predicate, AST statementList, boolean temporary) {
 	}
 	
 	public boolean execute(WordsEnvironment environment) throws WordsProgramException {
```
akallem: For clarity we should have Javadoc documentation here indicating that it returns true if it should be kept alive and false if it should be deleted.  (Edited my original comment where I said this backwards!)

src/words/ast/INodeListenerTemp.java
```diff
@@ -13,16 +13,25 @@ public ASTValue eval(WordsEnvironment environment) throws WordsRuntimeException
...
+			}
+			
+			return null;
 		}
-		
-		return null;
 	}
 }
```
akallem: It looks like these edits have only been made to INodeListenerTemp, but not INodeListenerPerm?  I think the two are identical except for the final boolean passed to evironment.createListener(), so perhaps they should be consolidated into one.

src/words/ast/INodeListenerTemp.java
alexthemark: I don't think that we want to actually have this eval here. I'm removing it -- and event listener should actually handle this case on its own.

akallem: OK, I'm fine with that.

src/words/ast/INodeMovesPredicate.java
alexthemark: Refactored

src/words/ast/INodeMovesPredicate.java
alexthemark, line 37: I don't really want to have to pass up a list of <alias, object> pairs (so that touches would work as well) and the environment to not repeat a couple lines of code.

alexthemark:
Includes the refactor of names now, so this also closes #97

alexthemark:
If anybody has any more comments, leave them here, but otherwise I'm gonna go ahead and merge in about an hour so that everyone can start their tasks from today.

akallem:
Before you do, would it be possible to incorporate James's while loop stuff?  That way everything is nice and tidy before we continue to the next set of issues.

akallem:
Don't merge yet, I'm getting a compile error when I try to build this branch.

akallem:
Looks like you didn't include LNodeBoolean in the commit.  Can you add it?

alexthemark:
Ah, my bad. Included now. I'll merge James's branch in as well.

akallem:
This is super cool and looks good to me.  I'm OK for this to be merged.

PR #104, opened 2015-04-18
Implement stop

### LINKED ISSUE #72: Implement Stop

youngthejames:
Closes #72

### DIFF
--- a/src/words/environment/WordsObject.java
+++ b/src/words/environment/WordsObject.java
@@ -22,7 +22,7 @@ public class WordsObject {
 		this.properties = new HashMap<String, Property>();
 	}
 	
-	public void clearQueue() {
+	public void clearActionQueue() {
 		actionQueue.clear();
 	}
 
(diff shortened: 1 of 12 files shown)


youngthejames:
Not sure if my JUnit test is particularly useful but I just wanted to make sure the method I added to WordsObject actually clears the queue.

akallem:
Looks good to me.

akallem:
Actually, something is wrong.  If I just open a Words console, create Fred, and then Stop Fred, I get an invalid type exception.  Looks like it's because you're assuming the reference list must be populated, but it's allowed to be empty here.  Take a look at INodeQueueMove and see how it handles the reference list; you should do the same here.

src/words/environment/WordsObject.java
```diff
@@ -21,6 +21,10 @@ public WordsObject(String objectName, WordsClass wordsClass, Position cell) {
 		this.actionQueue = new LinkedList<Action>();
 		this.properties = new HashMap<String, Property>();
 	}
+	
+	public void clearQueue() {
```
alexthemark: This name could be clearer (i.e. clearActionQueue), since an object could, in the future, have more than one queue.

akallem:
Cool, looks good to me now.

mzbenami:
cool!

zwdnyc:
looks great.

PR #105, opened 2015-04-19
More GUI Improvements

### LINKED ISSUE #103: String Escapes
It looks like the \n escape may not be rendering correctly in the GUI.  The other escapes seems to be OK but should be checked.

akallem:
This branch includes a bunch of simple improvements to the GUI:
- Each object is now rendered in a different random color, which makes it look more interesting.  The color is based on a hash of the object and class name, so it will be the same on every run and no matter where the object moves.
- As the GUI zooms in and out, the text size adjusts accordingly.
- If multiple objects are on the same spot, it renders them as cascading overlapping squares.  You can only see the top one, but at least it's clearer that there are other objects there too.
- Ensures newlines in the object message are rendered correctly

Closes #103


alexthemark:
Cool, I like a lot of it. 

2 Items of note:
- If I create over 10 objects in the same position, they start to flow out of the correct cell and into other cells.
- If I create over 155 objects in the same position, the entire program crashes.

akallem:
Good catches.  I modified with a simple maximum of 6 overlapping objects displayed per cell.  If there are more than 6, it'll just look like 6.  (In practice I doubt this will come up anyway.)

alexthemark:
lgtm now.

mzbenami:
awesome! thank you.

PR #106, opened 2015-04-19
Refactor Direction Class

akallem:
This branch does a slight refactor of the Direction class.  Previously, a Direction was an object that could be instantiated.  This actually could create some subtle problems when we flip a direction in a move because the value was negative, but the original LNodeDirection still has a handle to it.  (I encountered this as a subtle bug when experimenting with a Marco Polo sample program).  That could be fixed by always copying the Direction object when creating a new one, but I think the approach on this branch is safer and clearer.  Now, there is no Direction object per se.  The Direction class simply provides a static enum with the different directions.

This branch also has a minor unrelated bug fix in WordsObject.setProperty(), where after we set the row or column, it should exit, whereas previously is would go on to actually set a property called row or column.  (It never had any adverse effect, because the getProperty() method would always retrieve the current row or column.)

### DIFF
--- a/src/words/Words.y
+++ b/src/words/Words.y
@@ -302,11 +302,11 @@ queue_assign_property:
 	;
 
 direction:
-		ANYWHERE									{ $$ = new LNodeDirection(Direction.Type.ANYWHERE); ((AST) $$).lineNumber = lexer.lineNumber; }
-	|	DOWN										{ $$ = new LNodeDirection(Direction.Type.DOWN); ((AST) $$).lineNumber = lexer.lineNumber; }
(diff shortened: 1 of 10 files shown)


alexthemark:
I like having this static, but I think Java actually provides a better way of doing the same thing. Check out #107  which is a PR into this branch.

alexthemark:
Looks good to me. I'm in favor of merging this ASAP.

akallem:
Me too.  @mzbenami @ZhangWangda @jtyoung1292 can you take a quick look?

mzbenami:
slick, i didn't know you could use enums like that either.

zwdnyc:
looks great.

PR #107, opened 2015-04-19
Refactor Direction to not use Type

alexthemark:
Branched off Andrew's refactor of direction. By making "Direction" itself defined as an enum, we can get rid of that nasty ".Type" that had to be called every time we used direction. 

Note: This PR will merge into refactor-direction, not master. Then, once this is on refactor-direction (assuming you like it), we would merge Andrew's existing PR of refactor-direction into master.

### DIFF
--- a/src/words/Words.y
+++ b/src/words/Words.y
@@ -302,11 +302,11 @@ queue_assign_property:
 	;
 
 direction:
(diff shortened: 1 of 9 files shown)


akallem:
This is great!  I never knew you could declare an enum directly and still have methods and so forth in it.  I'm going to merge this now into refactor-direction, and then everyone else can take a look before we merge into master.

PR #111, opened 2015-04-21
Remove

mzbenami:
Object removal seeming pretty slick. I included a system-level test but omitted the log output because every frame was blank, because removal happens before the first frame. Unless there's a way to queue removal on the present code base?

### DIFF
--- a/src/words/environment/Environment.java
+++ b/src/words/environment/Environment.java
@@ -135,6 +135,24 @@ public class Environment {
 		}
 	}
 	
+	public void removeObject(WordsObject object) {
+		object.clearReferers();
+		
+		for (HashMap<String, WordsObject> scope : objectsByName) {
+			if (scope.get(object.getObjectName()) != null) {
+				scope.remove(object.getObjectName());
+				break;
+			}
+		}
+		
+		for (WordsClass wordsClass : objectsByClass.keySet()) {
+			HashSet<WordsObject> objectSet = objectsByClass.get(wordsClass);
+			if (objectSet.contains(object)) {
+				objectSet.remove(object);
+			}
+		}
+	}
+	
 	/**
 	 * Adds a named object to the most local scope in use. 
 	 */
--- a/src/words/environment/Environment.java
+++ b/src/words/environment/Environment.java
@@ -139,18 +139,12 @@ public class Environment {
 		object.clearReferers();
 		
 		for (HashMap<String, WordsObject> scope : objectsByName) {
-			if (scope.get(object.getObjectName()) != null) {
-				scope.remove(object.getObjectName());
+			if (scope.remove(object.getObjectName()) != null) {
 				break;
 			}
 		}
 		
-		for (WordsClass wordsClass : objectsByClass.keySet()) {
-			HashSet<WordsObject> objectSet = objectsByClass.get(wordsClass);
-			if (objectSet.contains(object)) {
-				objectSet.remove(object);
-			}
-		}
+		objectsByClass.get(object.getWordsClass()).remove(object);
 	}
 	
 	/**
--- a/src/words/environment/WordsObject.java
+++ b/src/words/environment/WordsObject.java
@@ -116,10 +116,6 @@ public class WordsObject {
 		}
 	}
 	
-	public void addReferer(WordsObject referer, Property refererProperty) {
-		
-	}
-
 	public void moveUp() {
 		this.cell.y++;
 	}
--- a/src/words/environment/WordsObject.java
+++ b/src/words/environment/WordsObject.java
@@ -17,26 +17,6 @@ public class WordsObject {
 	private Action lastAction;
 	private HashMap<WordsObject, ArrayList<Property>> referers;
 	
-	public void updateReferers(WordsObject referringObject, Property referringProperty) {
-		ArrayList<Property> referringProperties;
-		if ((referringProperties = referers.get(referringObject)) == null) {
-			referringProperties = new ArrayList<Property>();
-			referers.put(referringObject, referringProperties);
-		}
-		
-		referringProperties.add(referringProperty);
-	}
-	
-	public void clearReferers() {
-		for (WordsObject referringObject : referers.keySet()) {
-			for (Property property : referers.get(referringObject)) {
-				property.type = PropertyType.NOTHING;
-				property.objProperty = null;
-			}
-			referers.remove(referringObject);
-		}
-	}
-	
 	public WordsObject(String objectName, WordsClass wordsClass, Position cell) {
 		this.wordsClass = wordsClass;
 		this.objectName = objectName;
@@ -113,6 +93,26 @@ public class WordsObject {
 				property.objProperty.updateReferers(this, property);
 			}
 			properties.put(propertyName, property);
+		}
+	}
+	
+	public void updateReferers(WordsObject referringObject, Property referringProperty) {
+		ArrayList<Property> referringProperties;
+		if ((referringProperties = referers.get(referringObject)) == null) {
+			referringProperties = new ArrayList<Property>();
+			referers.put(referringObject, referringProperties);
+		}
+		
+		referringProperties.add(referringProperty);
+	}
+	
+	public void clearReferers() {
+		for (WordsObject referringObject : referers.keySet()) {
+			for (Property property : referers.get(referringObject)) {
+				property.type = PropertyType.NOTHING;
+				property.objProperty = null;
+			}
+			referers.remove(referringObject);
 		}
 	}
 	
--- a/src/words/environment/WordsObject.java
+++ b/src/words/environment/WordsObject.java
@@ -2,7 +2,10 @@ package words.environment;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedList;
+import java.util.Map;
+import java.util.Set;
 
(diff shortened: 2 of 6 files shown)


src/words/environment/Environment.java
```diff
@@ -135,6 +135,24 @@ public WordsObject createObject(String objectName, String className, Position po
 		}
 	}
 	
+	public void removeObject(WordsObject object) {
+		object.clearReferers();
+		
+		for (HashMap<String, WordsObject> scope : objectsByName) {
+			if (scope.get(object.getObjectName()) != null) {
```
akallem: I think technically this check is not necessary, since the remove() method says it removes a key if it exists and does nothing if not.

src/words/environment/Environment.java
```diff
@@ -135,6 +135,24 @@ public WordsObject createObject(String objectName, String className, Position po
...
+				scope.remove(object.getObjectName());
+				break;
+			}
+		}
+		
+		for (WordsClass wordsClass : objectsByClass.keySet()) {
+			HashSet<WordsObject> objectSet = objectsByClass.get(wordsClass);
+			if (objectSet.contains(object)) {
```
akallem: The same would apply here; I think this check is actually unnecessary.

src/words/environment/Environment.java
```diff
@@ -135,6 +135,24 @@ public WordsObject createObject(String objectName, String className, Position po
...
+		object.clearReferers();
+		
+		for (HashMap<String, WordsObject> scope : objectsByName) {
+			if (scope.get(object.getObjectName()) != null) {
+				scope.remove(object.getObjectName());
+				break;
+			}
+		}
+		
+		for (WordsClass wordsClass : objectsByClass.keySet()) {
```
akallem: Does this really need to be done as a loop?  objectsByClass is a HashMap from WordsClass to the HashSet.  And the WordsObject knows it's WordsClass.  So couldn't be just use the HashMap to lookup the appropriate HashSet, and then remove the WordsObject from that HashSet immediately?

src/words/environment/WordsObject.java
```diff
@@ -13,13 +15,35 @@
 	private Position cell;
 	private String currentMessage;
 	private Action lastAction;
-
+	private HashMap<WordsObject, ArrayList<Property>> referers;
```
akallem, line 21: I guess this could really be a LinkedList or even just a Set, since the order is not relevant.

src/words/environment/WordsObject.java
```diff
@@ -84,8 +108,16 @@ public void setProperty(String propertyName, Property property) throws WordsRunt
...
-		else
+		else {
+			if (property.type == PropertyType.OBJECT) {
+				property.objProperty.updateReferers(this, property);
+			}
 			properties.put(propertyName, property);
+		}
+	}
+	
+	public void addReferer(WordsObject referer, Property refererProperty) {
```
akallem: Looks like this blank method is unused and can be deleted.

src/words/environment/WordsObject.java
```diff
@@ -84,8 +108,16 @@ public void setProperty(String propertyName, Property property) throws WordsRunt
 
 		if (properties.containsKey(propertyName) && property.type == Property.PropertyType.NOTHING)
 			properties.remove(propertyName);
```
akallem, line 93: If I remove a property on an object, and that property was referring to another object, then I'm no longer a referrer to that object, and that object's referrer list would need to remove me as a referrer, right?

src/words/environment/WordsObject.java
```diff
@@ -13,13 +15,35 @@
 	private Position cell;
 	private String currentMessage;
 	private Action lastAction;
-
+	private HashMap<WordsObject, ArrayList<Property>> referers;
+	
+	public void updateReferers(WordsObject referringObject, Property referringProperty) {
+		ArrayList<Property> referringProperties;
+		if ((referringProperties = referers.get(referringObject)) == null) {
```
akallem: This is so C-like, it confused me for a bit, but I get what you're doing now.  :-)

src/words/environment/WordsObject.java
```diff
@@ -13,13 +15,35 @@
...
+		referringProperties.add(referringProperty);
+	}
+	
+	public void clearReferers() {
+		for (WordsObject referringObject : referers.keySet()) {
+			for (Property property : referers.get(referringObject)) {
+				property.type = PropertyType.NOTHING;
+				property.objProperty = null;
+			}
+			referers.remove(referringObject);
```
akallem: Since we're modifying the list of referers while we're in the loop, I think we need to use the iterator syntax rather than this "enhanced for loop" shorthand.  You can see how Wangda did it in event listeners in the frame loop.  See this post for details: http://stackoverflow.com/questions/1196586/calling-remove-in-foreach-loop-in-java.

src/words/environment/Environment.java
mzbenami: good call, check out how i'm doing it now. now i just check wether remove() returns null, and break if it doesn't. we should still break at the first scope at which the object is found, right?

src/words/environment/Environment.java
mzbenami: yes got rid of this whole loop.

src/words/environment/Environment.java
mzbenami: you're right once again. i changed this.

src/words/environment/WordsObject.java
mzbenami, line 21: not sure i follow here. do you mean a Set of WordsObjects, a Set of Propertys, or a Set of a new class that combines the two? I'll comment more on my thinking behind my implementation in the note below about what if an object removes a property that refers to this object.

src/words/environment/WordsObject.java
mzbenami: yea lazy on my part, my bad. it's gone now.

src/words/environment/WordsObject.java
mzbenami, line 93: This is actually what I consider the slickest part of my implementation. Your concern is valid, in addition to the concern if object A's property X refers object B, and then we set object A's property X to something else, shouldn't object A be removed as a referer to object B?

This is where our property type of "NOTHING" comes in to save the day. Since we decided not having a property assigned is the same has having NOTHING assigned, the clearReferers() method in the WordsObject class simply goes through all the Property objects and converts them to NOTHING. So if object A's X was pointing at object B, and we remove B, A's X is now pointing at NOTHING. If A's X was pointing at B, and then A's X was removed or set to some other property, it is no longer pointing at object B. Nevertheless, the original property will still be converted to NOTHING in clearReferers().

src/words/environment/WordsObject.java
mzbenami: Disagree here. I know what you mean about removing from a collection while iterating through it. But I am not iterating through the HashMap technically. I'm iterating through a copy of the keys to the HashMap.

src/words/environment/WordsObject.java
mzbenami: sorry didn't mean to confuse. i just like this way of doing things.

src/words/environment/WordsObject.java
akallem, line 21: Sorry, I was referring specifically to the ArrayList part, not the entire thing.  An ArrayList provides more functionality than we need to store the Propertys, since there's no need to retrieve them by index or iterate through them in a specific order.  But it's certainly OK as is. :-)

src/words/environment/WordsObject.java
akallem: I'm not sure about that.  According to the specification, the keySet isn't a copy, it's backed by the original map, so changes to the map during iteration result in undefined behavior: https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#keySet()

src/words/environment/WordsObject.java
akallem, line 93: I see, that's pretty cool.  Just to make sure I get it, object B's "referers" is not really the _current_ set of object properties pointing to B but the set of _all_ object properties that ever at one time had a property pointing to B.  Some of those may no longer be attached to their original object (A), but that's OK, we'll set them to NOTHING anyway.  And then they'll get garbage collected by Java shortly after, because they get removed from "referers" so no one has a handle to them any more.

src/words/environment/WordsObject.java
mzbenami, line 21: Oh interesting, yea I just kind of automatically go to ArrayList as my default iterable collection. I guess it could be a LinkedList. Something about iterating through a Set doesn't sit well with me. I feel like it's not meant to be used for that. I don't know I could be wrong. But yea let's leave it.

src/words/environment/WordsObject.java
mzbenami, line 93: That's correct.

src/words/environment/WordsObject.java
mzbenami: when you're right, you're right. and you're right. fixed.

akallem:
OK, this looks good to me.  It seems like the system test log was inadvertently not committed, so I added it.  I also modified the system test slightly to use listeners so that the objects aren't removed immediately.  That way, the log actually has some changes in it that are observable.

youngthejames:
Very cool Mike. Looks great.

zwdnyc:
nice!

PR #112, opened 2015-04-22
Adjacency Predicates

akallem:
This branch plays with a new language construct where a listener can test if members of different classes are adjacent to one another (up down left right), not just on top of each other (the existing "touches" predicate).  We can discuss this at our next meeting.

Right now it only tests adjacency by one cell, but it could be extended to allow farther away interactions.

If we agree with this, the grammar in the reference manual will have to be updated accordingly.


zwdnyc:
interesting

PR #113, opened 2015-04-23
Object to string

### LINKED ISSUE #66: Coerce Object to String
Leaving this here so we can discuss it on Monday. 

I'm wondering if we should allow coercion of objects to strings (in which case they would be coerced to the object name). This would be helpful in the following example.

Janet is a person.
Alex is a person.
Alex's mother is Janet.
Make Alex say "My mother is " + Alex's mother

Alternatively, maybe we could add a "name" property to every object, which would return the object's name?

I.e.
Make Alex say "My mother is " + Alex's mother's name.

youngthejames:
Closes #66

### DIFF
--- a/src/words/ast/ASTValue.java
+++ b/src/words/ast/ASTValue.java
@@ -82,6 +82,10 @@ public class ASTValue {
 					this.stringValue = String.format("%f", numValue);
 					this.type = newType;
 				}
+				if (type == Type.OBJ) {
+					this.stringValue = objValue.getObjectName();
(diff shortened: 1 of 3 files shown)


youngthejames:
You can also now feed the Say command an object and it will output the name of the object.

akallem:
Awesome, looks good to me.  Let's remember when we merge this to update the reference manual accordingly.

zwdnyc:
looks nice

PR #119, opened 2015-04-24
Make touches only fire when the objects move to touch

alexthemark:
Andrew realized that our current definition of touches did not reconcile with the reference manual. When two objects touch, they should fire an event listener for "touches." If they have been standing on the same spot, they should not fire.

### DIFF
--- a/src/words/ast/INodeAdjacencyPredicate.java
+++ b/src/words/ast/INodeAdjacencyPredicate.java
@@ -38,7 +38,8 @@ public class INodeAdjacencyPredicate extends INodeBasicActionPredicate {
 		
 		for (WordsObject object1: objectsToCheck1) {
 			for (WordsObject object2: objectsToCheck2) {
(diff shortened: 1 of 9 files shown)


akallem:
Looks good to me.  One minor question: what if two objects are created next to each other in the same frame (or one is created next to an existing object in the same frame).  Even though they haven't moved yet, they have come to be adjacent in that frame, so I think the listener should fire.

alexthemark:
I think that should fire it. I just pushed a commit that updates that (along with updated test cases). I would recommend making a similar action for adjacency. 

EDIT: Forgot that adjacency was merged, made the change.

akallem:
OK, this looks good.  I do think it would be clearer if, instead of testing whether lastAction is null, we have a getter method on whether the object was created.  Right now the code implicitly relies on lastAction being null only when an object is created, never later.  If that's an invariant we mean to keep, then we should at the very least document it and better yet test it.

PR #120, opened 2015-04-24
Use a fake wait flag instead of an actual wait action after each object ...

### LINKED ISSUE #98: Object Creation Wait
Objects should not do anything for their first frame. This should not happen due to a wait action, but instead due to a lack of action.

alexthemark:
...is created

Closes #98 

The loop still needs to fast forward through a frame in each of our test cases because the behavior of the object is similar to that of a wait action -- wait listeners simply will not be triggered.


akallem:
Looks good to me.

PR #121, opened 2015-04-24
Test Adjacency Predicate

akallem:
This branch just includes new unit tests and system tests for the adjacency predicate that was recently added.


alexthemark:
lgtm

PR #122, opened 2015-04-24
Fix INodeEquals Bug

akallem:
I discovered a bug in INodeEquals where if two objects were being compared and they were not equal, it failed to return false.  This branch fixes that and includes system tests for it.


alexthemark:
Looks good. Merge ASAP

PR #123, opened 2015-04-24
Fix Argument Evaluation

akallem:
This branch fixes a bug in the evaluation of arguments to custom actions.  Previously, the arguments were incorrectly being evaluated _after_ we have already moved into the scope of the custom action definition.  Instead, the arguments need to be evaluated in the scope of the caller, and then the _values_ have to be passed and installed in the custom action definition's scope.

I've modified the system test to catch this case in the future.


alexthemark:
Good catch, fan of merging ASAP

PR #124, opened 2015-04-24
Additional Pronouns

akallem:
This branch fixes a minor issue with the pronouns that can serve as references (his, her, its, their).  Previously these pronouns only had lowercase forms, but they also need to have capitalized forms because they can appear as the first token in a statement (e.g., in a custom action definition, someone could write "Their property is 3.").

There was a also a bug in the lexer where the pronouns had lower priority than identifiers and so were not being tokenized correctly.


alexthemark:
Looks good.

PR #126, opened 2015-04-24
GUI Left Right Flip

akallem:
Previously pressing left/right in the GUI would scroll in the opposite direction of what I think is intuitive (and opposite conceptually of how the up/down buttons worked).


alexthemark:
Good

PR #127, opened 2015-04-24
Tabs in Syntax Errors

akallem:
If a line had tabs, the ^ symbol in the syntax error wouldn't be lined up right.  This fixes that be replacing the tabs with spaces when reproducing the offending line.


alexthemark:
Good

PR #128, opened 2015-04-25
name and class properties added

### LINKED ISSUE #100: Access more properties of an object
Object's name and Object's class should be string properties that you can receive from an object.

youngthejames:
You can now access the Object's name with the property "name" and the Object's class with the property "class"

Closes #100

### DIFF
--- a/src/words/environment/WordsObject.java
+++ b/src/words/environment/WordsObject.java
@@ -97,13 +97,10 @@ public class WordsObject {
 			return new Property(cell.x);
 		
 		// Special handling for Object name "name" and Class name "class"
-		if (propertyName.equals("name")) {
+		if (propertyName.equals("name"))
 			return new Property(objectName);
-		}
-		
-		if (propertyName.equals("class")) {
+		else if (propertyName.equals("class"))
 			return new Property(wordsClass.getClassName());
-		}
 
 		Property property = getOwnProperty(propertyName);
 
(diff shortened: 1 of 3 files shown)


alexthemark:
I think these would be better handled in the same `if...else if` block as the other special cases (row and column). 

The set property method should also prevent people from attempting to set an object's name or an object's class. We don't allow people to change an object name or class after an object has been created, so we shouldn't let them think they can by setting a proeprty.

alexthemark:
These should also be unit tested. They're pretty simple so it's not essential, but the tests should be really simple to write, and it'll allow us to have better test coverage.

akallem:
Agree with Alex on both of this comments.

youngthejames:
Do you think I should add another exception to handle the user trying to modify name or class or should I leave it as is?

akallem:
It would have to be a new exception.  Maybe call it ImmutablePropertyException.

akallem:
Either that, or it just silently fails and doesn't update the property.  But I think using an exception is better because it tells the user they tried to do something that isn't allow.

youngthejames:
new error is called ModifyObjectPropertyException

akallem:
Cool, that's a good name.

src/words/environment/WordsObject.java
```diff
@@ -95,6 +95,12 @@ public Property getProperty(String propertyName) {
 			return new Property(cell.y);
 		else if (propertyName.equals("column"))
 			return new Property(cell.x);
+		
+		// Special handling for Object name "name" and Class name "class"
+		if (propertyName.equals("name"))
```
akallem: As Alex suggested, this should just be a continuation of the above if (as another else if) rather than a new if.

akallem:
Looks pretty good; see my comment about combining your if block with the above if block.  I'd also suggest a unit test that getting the name and class name work.

akallem:
Cool, this looks good to me.  I'll preemptively update the reference manual for this.

PR #129, opened 2015-04-25
Fix Position Creation

akallem:
This branch fixes a bug in where objects are placed.

For an object creation statement like:
  Fred is a thing at 1,3.
the reference manual says 1 is the _row_ and 3 is the _column_.  Our Position class stores positions in terms of x and y, which means the row should be y and the column should be x.  However, INodePosition was creating Positions in reverse, where the row was assigned to x and the column to y.

The fix was simple, and the updates to the tests just involved swapping the values.  (Often, our tests use the same row and column, which is part of the reason why we hadn't noticed.  The other is that we may intuitively be thinking of positions as x,y, not row,column.)

### DIFF
--- a/src/words/ast/INodePosition.java
+++ b/src/words/ast/INodePosition.java
@@ -21,6 +21,6 @@ public class INodePosition extends INode {
 			throw new InvalidTypeException(ASTValue.Type.NUM.toString(), col.type.toString());
 		}
 		
-		return new ASTValue(new Position(row.numValue, col.numValue));
+		return new ASTValue(new Position(col.numValue, row.numValue));
 	}
 }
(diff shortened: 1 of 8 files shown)


alexthemark:
I'm not convinced that we should be changing the code and not the reference manual. If we've all missed it because we've been doing the intuitive thing, maybe our language shouldn't do just the opposite.

akallem:
I'd be fine with that too.  I think the right solution would be to create objects at x,y (as in the existing implementation, and change the reference manual), but then also create special x,y properties similar to the special row column properties.  Otherwise, it's odd that a user would create an object at x,y but then retrieve properties as row column and then have to flip them again if creating a new object based on another object's position.

PR #136, opened 2015-04-27
Speed Option

akallem:
Added a simple option to control the speed.
  java -jar jar/Words.jar -s 100 program.words

The integer following -s is the number of milliseconds per frame.  If it's not an integer, it's ignored.


alexthemark:
looks good to me.

youngthejames:
nice!

PR #142, opened 2015-05-08
Predicate move fix

### LINKED ISSUE #139: Adjacency/Touches Predicate After Position Set
If an object's position is set directly with the row/column properties, it should fire the touches listener even if it's last action was not a move.

alexthemark:
The system test ListenerTest.Words has been modified to also test this situation. 

This works for both immediate updates (i.e. "Fred's row is 3.") and queued updates (i.e. "Make Fred's row be 3.") The difference was not quite trivial, since queued updates need to check current/past position after the execution of an action, while immediate assignments happen before an action is executed. 

Closes #139


akallem:
Cool, I'm taking a look now.  Not to be too much of a stickler, but we should probably call it movedInLastFrame rather than movedInLastTurn, since "frame" is our "official" word for it.

src/words/environment/WordsObject.java
```diff
@@ -198,8 +200,13 @@ public void moveRight() {
 	public void executeNextAction(Environment environment) throws WordsProgramException {
 		if (createdInThisFrame) {
 			createdInThisFrame = false;
-			return;
 		} else {
+			if (!lastCell.equals(cell)) {
+				this.movedInTurn = true;
+			}
+			else {
```
akallem: Not following our style guide here.  :-)

src/words/environment/WordsObject.java
```diff
@@ -216,11 +223,21 @@ public void executeNextAction(Environment environment) throws WordsProgramExcept
...
 				action.execute(this, environment);
+				if (!lastCell.equals(cell)) {
+					this.movedInTurn = true;
+				}
 			} else {
 				lastAction = new WaitAction(environment.getCurrentScope());
 			}
 		}
+		// Store the last position based on actions
+		lastCell = new Position(cell);
```
akallem: On the surface, this line is weird to me.  It sets lastCell to be cell, which means they are the same most of the time (except for a short period inside this method).  What's the reason it's here?

akallem:
I find the implementation a bit confusing.  It's not clear exactly what invariant "lastCell" is obeying.

Could I suggest something maybe simpler?  Before phase 1, just have all objects set lastCell = cell.  Then have the movedInLastTurn method (which we could rename to movedInLastFrame) just check if lastAction is null or if lastCell != cell.

In this scheme, executeNextAction() wouldn't even need to be modified.

Does that make sense?

alexthemark:
That is a lot simpler... I'll revise this.

Sorry, my brain has been fried this week.

akallem:
Cool, this is looking good now.  Now that I think of it, technically the method should probably be called movedThisFrame() rather than movedInLastFrame(), since it's used within a frame, not across frames.  And for clarity, lastRecordedPosition could be called startOfFramePosition (though I think that's less important).

But otherwise it looks good to me.

mzbenami:
lgtm, especially love the commit messages

PR #146, opened 2015-06-25
Local variable

alexthemark:
Just because the class is over doesn't mean that we can't complete a few more features, right? I had talked to Andrew about adding local variables, and we both thought that it would really complete the language. This means that we can now write programs like this:

Alex is a thing at 0,0.
variable is 5.
Alex's property is variable.
Make Alex say Alex's property.
variable is "hello".
Make Alex say Alex's property.

It's important to note that variables are mutable, so they can be overwritten (much like an object's properties). Additionally, programs refer to variables by reference and not by value. Passing variables around by reference was actually harder than I had expected: in our past structure we had the `Property` class in the Environment, and `ASTValue` in the AST. Passing information between these two classes, we often lost references to the original objects. To fix this, I combined the classes into a new `Variable` class that's used in both the environment and the AST (I think this is also an improvement overall -- the two classes were largely copies of each other). 

The only issue that I see with my current implementation is that we can't queue an assignment onto a local variable (i.e. "Make variable be 7.") since the variable doesn't have an action queue. Queued assignment seems like an important feature, since it's the only way to change a local variable over time.To fix this, I could refactor variables into an inherited structure (i.e. WordsObject extends Variable), give all variables an action queue, and then allow different subclasses to enqueue different actions (i.e. everything can enqueue assignment, but only WordsObject can enqueue move, say, etc.), but that's a pretty big change. Was wondering if anyone had any thoughts on how else to do this or the current implementation.... (Or you could ignore this PR since the class is over). 

P.S. How y'all doing?


akallem:
First of all, let me say: yeeeessss!!!!  I've been waiting for / expecting this PR, and I'm quite excited to see it!  :-)

Before I dive in, I wanted to clarify an interesting comment you made, which is that variables are all passed by reference(?).  So in this example:

A Tester is a thing which {
  can make_a_change with variable which means {
    variable is variable+1.
  }
}

Test is a Tester at 0,0.
y is 1.
Make Test modify_a_variable with variable y.
Make Test say y.

What's the expected/intended behavior?  Test will eventually say whatever "y" evaluates to at the time the say action is executed.  But first, the custom action modify_a_variable was executed.  Did that custom action actually modify y, making it 2, because y was passed by reference?

If so, I think that would be confusing and unlike most languages where numbers and strings are passed by value, which then would then make it not very educational or a "stepping stone" to those languages.

I guess I could ask the same about the example you gave.  What's the expected/intended behavior?  Given that the RHS of the say action isn't evaluated until after all those statements are executed, I could see both say statements either saying "5" (if "Alex's property is variable" assigns by value) or both saying "hello" (if "Alex's property is variable." makes "property" point to "variable", i.e. assign by reference.)  But I don't think it'll say "5" followed by "hello".

Also, can you give an example of what you mean when you said passing information between the old Property and ASTValue classes could often(?) cause us to lose references to the original(?) objects?  I'm not sure which objects you're referring to (the "passer" Property or ASTValue?  a WordsObject?), or why it matters.

I do agree that Property and ASTValue are quite similar, and there might be a case for combining them as you did.  On the other hand, I think there are some differences too.  First, in the language right now, a Property can never be, say, a direction (although perhaps we could extend the language to allow that).  So combining the classes would confuse that.  Second, in the language right now, a Property is a persistent object that exists across frames, whereas an ASTValue is just a crutch used during AST evaluation but never existing beyond/outside that context.

P.S.  I'm doing quite well, how are you all?!

alexthemark:
The way I wrote my example is actually a bit ambiguous. Since it contains two immediate statements, `variable` would actually be set to "hello" almost immediately, and Alex would probably only say "hello" (twice). If you waited for Alex to say "5", then modified variable to be "hello", then made Alex say his property again, then he would say "hello".

In your example, y would indeed be 2 by the time Test says y. Truth be told, custom function parameters had slipped my mind a bit here. I was thinking about how objects store their properties -- which I definitely think should be by reference. I think that, for functions, it would be completely reasonable to pass by value, C style. Commit c73e169 will actually pass by value (if you include the file I forgot to check in until 06a0132). 

Here's an example of passing information between Property and ASTValue where a reference is lost.

Alex is a thing at 0,0.
variable is 4.
Alex's age is variable. 

When I set Alex's property to "variable", the Property variable is fetched by INodeRetrieveProperty and is then placed into an ASTValue of type num. Then, at INodeAssignProperty, the ASTValue is translated into a new Property object held in Alex's properties table. Now, there are two different property objects, both of value 4. If I modify Alex's age, it doesn't modify variable, and if I modify variable, it doesn't modify Alex's age. By keeping one consistent (java) object that is fetched and then assigned, references are maintained.

All of this is to facilitate passing by reference, which is up for debate (since I didn't really talk it out or consult anyone on it). I'm about to write a separate post on why I passed by reference (just to keep things to a reasonable length).

alexthemark:
Actually I can't justify it at all. I think I should revert it to passing by value. PR #147 passes things by value.

Sure was fun to program though.

alexthemark:
The Words variables that never were....

PR #147, opened 2015-06-26
Pass by value

alexthemark:
Very similar to the other PR... except this one passes by value. Note that there are no test cases written yet, but I wanted to give you an idea of this as well. :)


src/words/ast/INodeCreateLocalVariable.java
```diff
@@ -0,0 +1,28 @@
...
+		ASTValue varName = children.get(0).eval(environment);
+		ASTValue rawValue = children.get(1).eval(environment);
+		Variable value = null;
+		
+		if (rawValue.type == ASTValue.Type.NUM) {
+			value = new Variable(rawValue.numValue);
+		} else if (rawValue.type == ASTValue.Type.STRING) {
+			value = new Variable(rawValue.stringValue);
+		}
+
```
akallem, line 24: What about handling the case where the ASTValue is NOTHING?  It should be legal to assign a variable to be nothing, as in, "x is nothing.".

Also, is it legal to assign a variable to an object?  For example:
  Fred is a person.
  X is Fred.
(If that's legal, garbage collection becomes harder, so it's probably not legal.  But if it's not legal, then the code here will silently do nothing, whereas I think it should give a runtime exception to the user.)

src/words/environment/Environment.java
```diff
@@ -127,19 +127,32 @@ public int getScopeDepth() {
...
+			throw new ObjectAlreadyExistsException(varName);
+		} else if (existingVariable.type != Variable.VariableType.NOTHING) {
+			existingVariable.copyOtherVariable(value);
 		} else {
-			throw new ObjectAlreadyExistsException(objectName);
+			getCurrentScope().variables.put(varName, value);
+		}
+		
+		// If this variable is an object, also add it to objects by class
+		if (value.type == Variable.VariableType.OBJECT) {
```
akallem, line 154: This will throw a null pointer exception if value is null, and it would be null if INodeCreateLocalVariable is parsing a statement like "x is nothing" or "x is Fred".

alexthemark:
I just restricted INodeCreateLocalVariable so that local variables can only be numbers and strings (should have done this from the get-go). Let me know your current thoughts. If things seem good, I'll add tests and update our docs.

akallem:
Sorry for the delay, I was away for the weekend.  I just checked it out, and everything is looking good, so yes, go ahead and write the tests and update the documentation.  Good stuff!

alexthemark:
Sorry that took a while -- I've had a busy time moving in to my new apartment for the last week. :)

There are not many new tests. The existing object creation in TestEnvironment actually tests a lot of the variable creation, scoping, etc. I probably could have gone through and renamed some things, or modified a bit so that not all the tests were on objects... but it seems adequate to me for something that we already handed in.

akallem:
Cool, seems good to me.

One question: should it be allowed to write, say, "x is nothing." where x is a variable (rather than an object)?  Currently that's not allowed, but perhaps it should be.

But certainly nothing to worry about for this PR.

Cool stuff!

alexthemark:
"nothing" to worry about, eh?

PR #148, opened 2015-07-26
Revise Turing Machine Sample

akallem:
This revision makes the Turing Machine sample more properly object-oriented, where concrete Turing Machines can inherit from a base class called 'head' that provides the common behaviors.  The sample now has two different Turing Machines which run in parallel.  It also uses the new variable functionality.


alexthemark:
ooh, so exciting. I've always wanted to have multiple turing machines running in Words. 

LGTM.

